PR #15, opened 2015-06-02
Allow easy VaadinServlet override

lightoze:
Sometimes it is desirable to add some custom logic to (Spring)VaadinServlet. This patch makes it super-easy.


lightoze:
Moved to Gerrit

PR #29, opened 2016-04-24
Translate application title (fixes #28)

### LINKED ISSUE #28: Make application title translatable
With the annontation `com.vaadin.annotations.Title` you can set the title of an Vaadin UI. The title text is used for the HTML page title.

The title of an Vaadin UI should be translatable by using the Spring message source of the application. Therefore an new annotation should be defined, that takes an message code for the message source as parameter. If there is no message for the message code defined, a default text containing the message code should be displayed. The guarantee the backwards compatibility the annontation `com.vaadin.annotations.Title` should be used if is defined (no matter if the new annotation is also defined in application code for the UI).

KyleRogers:
Hi,

what do you think about this fix? Would be glad if you could merge it in. Fixes #28 

If you don't like it but know another way to implement this, please elaborate on that approach so that we can consider to implement it and provide a new or updated pull request.

Regards,
Stefan

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/29)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/annotation/TranslatableTitle.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/annotation/TranslatableTitle.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2015 The original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


KyleRogers:
Migrated pull request to gerrit: https://dev.vaadin.com/review/#/c/13610/

hesara:
Reopening this and closing the Gerrit change as the project has been migrated to a fullt GitHub based workflow.

KyleRogers:
Replaced by #209 against branch 1.2

PR #32, opened 2016-07-21
Feature/fix serialization failure

khauser:
The way I came around the first initial serialization problems while persisting a vaadin session. Both SpringUIProvider and SpringViewProvider now get their application context from an external object.


markathomas:
The only problem with this approach is it assumes you're deploying a single Vaadin servlet to the servlet container.  To support multiple deployments you'd need to register the app context to the new class by something Serializable and accessible to the user classes.  Perhaps the session id?

khauser:
There is always time to learn.. Is there something analog I can look at? I have no clue about what you mean with "new class" and "user classes", but suspect you are right in what you say. Maybe you can explain it a little bit closer for me.
The pull request is also open for any changes you suggest.

Artur-:
This was fixed in 5a34f676e3ff997d0208785a56d1e6858e4139e4

PR #108, opened 2016-10-13
Make SpringViewProvider UI scoped

### LINKED ISSUE #37: SpringViewProvider should be UIScoped
`SpringViewProvider` is currently a singleton bean even though it ends up in each `VaadinSession`  and is thus serialized with the session. In practice after the fix for #35, it is not even a singleton bean but "most of the time maybe a singleton bean".

The view provider should really be UIScoped so that it tracks only the views for one view.

hesara:
This is logically more correct and allows more granular serialization.

Closes #37

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/108)


elmot:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/108)*

PR #117, opened 2016-10-14
Fixes false multiple ViewContainer detection

elmot:
Not sure if this is most optimal solution but it seems working

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/117)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java
@@ -15,48 +15,54 @@
  */
 package com.vaadin.spring.internal;
 
-import java.lang.reflect.Field;
-
+import com.vaadin.spring.annotation.ViewContainer;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.support.BeanDefinitionValidationException;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.ReflectionUtils.FieldCallback;
 
-import com.vaadin.spring.annotation.ViewContainer;
+import java.lang.reflect.Field;
+import java.util.WeakHashMap;
 
 public class ViewContainerPostProcessor implements BeanPostProcessor {
+    private WeakHashMap<Class<?>, Field> viewContainerFieldsByClass = new WeakHashMap<Class<?>, Field>();
     private Object viewContainer = null;
 
     @Override
-    public Object postProcessAfterInitialization(final Object bean,
-            String beanName) throws BeansException {
-        ReflectionUtils.doWithFields(bean.getClass(), new FieldCallback() {
-            @Override
-            public void doWith(Field field)
-                    throws IllegalArgumentException, IllegalAccessException {
-                if (field.isAnnotationPresent(ViewContainer.class)) {
-                    if (viewContainer == null) {
-                        field.setAccessible(true);
-                        viewContainer = field.get(bean);
-                        field.setAccessible(false);
-                    } else {
-                        throw new BeanDefinitionValidationException(
-                                "Multiple definitions of @"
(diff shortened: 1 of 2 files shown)


hesara:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java, line 29 at r1](https://reviewable.io:443/reviews/vaadin/spring/117#-KUGLyyk4PazOpib0_o4:-KUGLyyk4PazOpib0_o5:b4nqzic) ([raw file](https://github.com/vaadin/spring/blob/48f4ca1d162cbf14adda2fb59539960e4383d3f2/vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java#L29)):*
> ```Java
> 
> public class ViewContainerPostProcessor implements BeanPostProcessor {
>     private WeakHashMap<Class<?>, Field> viewContainerFieldsByClass = new WeakHashMap<Class<?>, Field>();
> ```

WeakReference and weak maps may cause problems for serialization in case this class is serialized e.g. between cluster nodes. Usually if this approach is used, the field needs to be transient and initialized whenever referenced and null. However, in this case it might be more complicated.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/spring/117#-KUGOOiSf2HTbTOQB8Wg:-KUGOOiSf2HTbTOQB8Wh:b-w71xug) ([raw file](https://github.com/vaadin/spring/blob/48f4ca1d162cbf14adda2fb59539960e4383d3f2/vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java#L30)):*
> ```Java
> public class ViewContainerPostProcessor implements BeanPostProcessor {
>     private WeakHashMap<Class<?>, Field> viewContainerFieldsByClass = new WeakHashMap<Class<?>, Field>();
>     private Object viewContainer = null;
> ```

actually this should really be by UI scope somehow rather than global (not a new issue of this patch, but the relates to the whole approach)

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java, line 45 at r1](https://reviewable.io:443/reviews/vaadin/spring/117#-KUGNZwRj940G-FSKTCA:-KUGNZwRj940G-FSKTCB:bl0myat) ([raw file](https://github.com/vaadin/spring/blob/48f4ca1d162cbf14adda2fb59539960e4383d3f2/vaadin-spring/src/main/java/com/vaadin/spring/internal/ViewContainerPostProcessor.java#L45)):*
> ```Java
> 
>     @Override
>     public Object postProcessBeforeInitialization(final Object bean, String beanName)
> ```

the "before" and "after" here are relative to a single bean, perhaps potentially leading to a situation where some viewcontainers are set up before a conflict is detected (although it seems that could happen before as well)

---

*[vaadin-spring/src/test/java/com/vaadin/spring/server/SpringUIProviderTestWithViewContainerOnField.java, line 60 at r1](https://reviewable.io:443/reviews/vaadin/spring/117#-KUGMh3XxSPxdM0bKo1e:-KUGMh3XxSPxdM0bKo1f:b-r08dfq) ([raw file](https://github.com/vaadin/spring/blob/48f4ca1d162cbf14adda2fb59539960e4383d3f2/vaadin-spring/src/test/java/com/vaadin/spring/server/SpringUIProviderTestWithViewContainerOnField.java#L60)):*
> ```Java
>     @SpringUI(path = "/next")
>     private static class TestUINext extends UI {
>         @ViewContainer
> ```

as another test if this also fixes that case (possibly not), @ViewContainer should be enabled for SpringViewProviderTest.TestUI2

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/117)*

hesara:
Superceded by #118. Estimate moved there in ZenHub.

PR #119, opened 2016-10-19
Switch to spring-boot:1.4.1.RELEASE and spring:4.3.3.RELEASE

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/119)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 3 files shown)


hesara:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[pom.xml, line 203 at r1](https://reviewable.io:443/reviews/vaadin/spring/119#-KURzBtkYPnNI6bRWh9s:-KURzBtkYPnNI6bRWh9t:by2ag1s) ([raw file](https://github.com/vaadin/spring/blob/c712c4d5b1d0d67ca68b7f46d764a84f67cce485/pom.xml#L203)):*
> ```XML
>         <module>vaadin-spring-boot-starter</module>
>     </modules>
>     <dependencyManagement>
> ```

this also leaks to user projects, whereas the old test dependencies didn't - it might be better to use a property in this pom rather than dependencyManagement

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/119)*

hesara:
Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/119)*

snicoll:
Does that mean that the next release will only work with Boot 1.4? I am pretty convinced it will work just fine with 1.3 but maybe we could have a separate build to ensure that? Switching to 1.4 for 1.1 is fine as well but start.spring.io should be aware of that.

PR #122, opened 2016-10-20
Support view activation listeners

### LINKED ISSUE #68: Scope vaadin-view not active in ViewChangeListener.afterViewChange
**Originally by _lorenzo.benvenuti_ on 2015-09-17 14:02:25 +0300**

___
Hi,

when using a ViewChangeListener the scope "vaadin-view" could be inactive when calling the afterViewChange method. I guess it depends on the order in which the ViewChangeListener are invoked: since the ViewBeanStore is a ViewChangeListener itself and it switches the active view on/off in the afterViewChange method, if a ViewChangeListener is invoked _between_ two ViewBeanStore there is no active view. I attach an Eclipse project that replicates the issue.

Thank you,

lorenzo

___
 Imported from https://dev.vaadin.com/ issue #18923

### LINKED ISSUE #80: ViewScope not properly initialized when navigating directly into the view
**Originally by _peterl1084_ on 2016-09-20 09:07:03 +0300**

___
I have a following view / component structure defined:

```
@SpringView(name="myview")
class MyView implements View {

  @Autowired
  private SomeComponent c;

  public void doSomething() {
      ...
  }
}
```

```
@ViewScope
class SomeComponent ... implements Component {

  @Autowired
  @Lazy
  private MyView view;

  ...

  public void someMethod() {
        view.doSomething();
   }
```

There is a view which @Autowires a @ViewScoped component that has a reference back to the view through a @Lazy proxy.

The above works fine and the doSomething method from the view can be invoked.

The problem is that if this view is navigated into directly (with URI fragment as the first view, without going through some other view first, the vaadin-view scope is not active for current thread.

this works:
```
localhost:8080 -> manually navigate into the view through some menu -> locahost:8080/#!myview
```

this does not work:
```
localhost:8080/#!myview
```

The thrown exception is:

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'myview': Scope 'vaadin-view' is not active for the current thread; ...; nested exception is java.lang.IllegalStateException: No active view

I assume that some listener that sets the scope / thread initially is not properly called at the beginning of the request. Debugging the navigator / view provider shows that the view bean exists but when the proxy reference back to the view is attempted to be resolved the 

```
com.vaadin.spring.internal.DefaultViewCache.getCurrentViewBeanStore()
```

tries to reference viewUnderConstruction variable, which is null.

The viewUnderConstruction is not null if view is navigated into through another initial view.

___
 Imported from https://dev.vaadin.com/ issue #20303

hesara:
This makes the view scope work properly by aligning events with when the effective scope actually changes.

This change requires the Navigator used to extend SpringNavigator to support view change events.

Fixes #68
Fixes #80 

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/122)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java
@@ -27,6 +27,7 @@ import org.springframework.util.Assert;
 
 import com.vaadin.navigator.Navigator;
 import com.vaadin.navigator.View;
+import com.vaadin.navigator.ViewChangeListener;
 import com.vaadin.spring.navigator.SpringNavigator;
 import com.vaadin.spring.navigator.ViewActivationListener;
 import com.vaadin.ui.UI;
@@ -55,21 +56,23 @@ public class DefaultViewCache implements ViewCache {
     @SuppressWarnings("serial")
     public DefaultViewCache() {
         Navigator navigator = getCurrentUI().getNavigator();
-        if (!(navigator instanceof SpringNavigator)) {
-            throw new IllegalStateException(
-                    "Navigator is not a SpringNavigator");
-        }
-        listener = new ViewActivationListener() {
-            @Override
-            public void viewActivated(ViewActivationEvent event) {
-                if (!event.isActivated()) {
-                    viewDeactivated(event.getViewName());
-                } else {
-                    DefaultViewCache.this.viewActivated(event.getViewName());
+        if (navigator instanceof SpringNavigator) {
+            listener = new ViewActivationListener() {
+                @Override
+                public void viewActivated(ViewActivationEvent event) {
+                    if (!event.isActivated()) {
+                        viewDeactivated(event.getViewName());
+                    } else {
+                        DefaultViewCache.this
+                                .viewActivated(event.getViewName());
+                    }
                 }
-            }
-        };
-        ((SpringNavigator) navigator).addViewActivationListener(listener);
+            };
+            ((SpringNavigator) navigator).addViewActivationListener(listener);
+        } else {
+            LOGGER.warn(
+                    "The Navigator used does not extend SpringNavigator. View scope support may be limited, and future versions of Vaadin Spring may remove support for using a plain Navigator.");
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


hesara:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java, line 129 at r2](https://reviewable.io:443/reviews/vaadin/spring/122#-KUX7_QP2ebDgCqBy3RF:-KUX7_QP2ebDgCqBy3RG:bs4mr87) ([raw file](https://github.com/vaadin/spring/blob/f6fad53b79c8daec42dce10ff69a0269bfa42991/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java#L129)):*
> ```Java
>             activeView = null;
>         }
>         getBeanStore(viewName).destroy();
> ```

deactivating a ui scoped view now fails because a view scope bean store has not been created for it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/122)*

hesara:
Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java, line 129 at r2](https://reviewable.io:443/reviews/vaadin/spring/122#-KUX7_QP2ebDgCqBy3RF:-KU_tH-EX9g0ap8aK_gC:bns63a3) ([raw file](https://github.com/vaadin/spring/blob/f6fad53b79c8daec42dce10ff69a0269bfa42991/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java#L129)):*
<details><summary><i>Previously, hesara (Henri Sara) wrote…</i></summary>
> deactivating a ui scoped view now fails because a view scope bean store has not been created for it
</details>
fixed in b3d24c3

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/122)*

elmot:
Reviewed 2 of 2 files at r3.
Review status: 2 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/122)*

elmot:
Reviewed 2 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/122)*

PR #133, opened 2016-10-25
Create new instance of view when re-navigating to it

### LINKED ISSUE #130: Same View instance reused when navigating to the same view
If a certain view is already navigated into and navigator.navigateTo is invoked that targets the same view (= URI will not change after the navigation) the view bean that SpringViewProvider returns to the caller is the same instance in which we currently are.

This could be by design in some cases but for example VaadinCDI Integration works differently and returns a new instance.

I would assume that VaadinSpring should work similarly or provide a configuration option on which behaviour is required.

hesara:
Fixes #130

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/133)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java
@@ -48,6 +48,9 @@ public class DefaultViewCache implements ViewCache {
     private Map<String, ViewBeanStore> beanStores = new HashMap<String, ViewBeanStore>();
 
     private String viewUnderConstruction = null;
+    // during the re-creation of a view with the same name, keep the old context
+    // in case the creation fails
+    private ViewBeanStore viewBeanStoreBeingReplaced = null;
 
     private String activeView = null;
 
@@ -86,6 +89,11 @@ public class DefaultViewCache implements ViewCache {
     @Override
     public void creatingView(String viewName) {
         LOGGER.trace("Creating view [{}] in cache [{}]", viewName, this);
+        ViewBeanStore oldBeanStore = beanStores.get(viewName);
+        if (oldBeanStore != null && viewName.equals(oldBeanStore.viewName)) {
+            viewBeanStoreBeingReplaced = oldBeanStore;
+            beanStores.remove(viewName);
+        }
         getOrCreateBeanStore(viewName);
         viewUnderConstruction = viewName;
     }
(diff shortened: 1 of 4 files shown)


hesara:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/spring/133#-KUw1hOvQ50wFmnhiBWV:-KUw1hOvQ50wFmnhiBWW:bemnrzz) ([raw file](https://github.com/vaadin/spring/blob/cdad55baca4080821ddd8de00c3b4c2d8f3b0371/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java#L113)):*
> ```Java
>      */
>     @Override
>     public void viewCreated(String viewName, View viewInstance) {
> ```

Forgetting the old context already here is actually problematic, as the actual scope change only happens later from Navigator and before that we get a ViewActivationEvent for the deactivation of the old view. Thus, after creation, we should go back to the old scope and be ready to actually activate the new one only when the actual scope change occurs (SpringNavigator.switchView()), assuming it is still valid.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/133)*

hesara:
Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, all discussions resolved.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/spring/133#-KUw1hOvQ50wFmnhiBWV:-KUw7xlchPj-rmy0p18_:b-1dws6v) ([raw file](https://github.com/vaadin/spring/blob/cdad55baca4080821ddd8de00c3b4c2d8f3b0371/vaadin-spring/src/main/java/com/vaadin/spring/internal/DefaultViewCache.java#L113)):*
<details><summary><i>Previously, hesara (Henri Sara) wrote…</i></summary>
> Forgetting the old context already here is actually problematic, as the actual scope change only happens later from Navigator and before that we get a ViewActivationEvent for the deactivation of the old view. Thus, after creation, we should go back to the old scope and be ready to actually activate the new one only when the actual scope change occurs (SpringNavigator.switchView()), assuming it is still valid.
</details>
Addressed in r2

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/133)*

elmot:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/133)*

Issue #151, opened 2016-11-10
Cannot use auto-configuration without com.vaadin.spring.navigator.SpringNavigator


snicoll:
`VaadinAutoConfiguration` has a constructs where it checks if a bean of a certain type is missing and if that's the case, auto-configures one. What is particular about this example is that it uses a `String` literal (rather than a `Class` literal) to references the target type.

You can't really do that without some more guards. That condition is going to inspect the application context for a bean of that type and will have to load the class in order to do so. I am happy to offer a PR if you want me to.

You can reproduce by upgrading the starter in the [`spring-data-vaadin-crud` sample project](https://github.com/mstahv/spring-data-vaadin-crud).

mstahv:
Dough :-( It still worked with rc2. I'll try to find the team to create a 1.0.1 really soon with this fix.

PR #154, opened 2016-11-10
Properly guard absence of SpringNavigator

snicoll:
Since `SpringNavigator` may not be present, it should be properly guarded
with an additional `ConditionalOnType` at class level. That way, the whole
processing of the `@Configuration` class will be skipped.

To avoid `SpringViewDisplayPostProcessor` to not be registered in such a
case, the registration of `SpringNavigator` has been moved to its own
configuration class.

Closes gh-151

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/154)


hesara:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/154)*

PR #157, opened 2016-11-21
Support serialisation in 1.1

mstahv:
1.1 release introduced a regression regarding session serialisation. There is 
a new direct reference to ApplicationContext. This pull request fixes
that in the same way as @Artur- fixed the previous serialization issue for 1.0.x release.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/157)

### DIFF
--- a/vaadin-spring-boot/src/main/java/com/vaadin/spring/boot/VaadinAutoConfiguration.java
+++ b/vaadin-spring-boot/src/main/java/com/vaadin/spring/boot/VaadinAutoConfiguration.java
@@ -29,6 +29,8 @@ import com.vaadin.spring.annotation.UIScope;
 import com.vaadin.spring.boot.annotation.EnableVaadinServlet;
 import com.vaadin.spring.internal.SpringViewDisplayPostProcessor;
 import com.vaadin.spring.navigator.SpringNavigator;
+import com.vaadin.spring.navigator.SpringViewProvider;
+import org.springframework.context.ApplicationContext;
 
 /**
  * @author Petter Holmström (petter@vaadin.com)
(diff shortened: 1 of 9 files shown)


macjuan:
Let's be sure that we'll have a automated test for this in the future.

hesara:
one API incompatibility and no test, but the API incompatibility can be refactored out (resulting in a smaller patch)

---

Reviewed 3 of 9 files at r1.
Review status: 3 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-spring/src/main/java/com/vaadin/spring/navigator/SpringNavigator.java, line 65 at r1](https://reviewable.io:443/reviews/vaadin/spring/157#-KX69mKCO5xzHNdr0aDv:-KX69mKCO5xzHNdr0aDw:bcf0pul) ([raw file](https://github.com/vaadin/spring/blob/e5a27b0ff33d3ae29b7e19b8bde5895f86df16f8/vaadin-spring/src/main/java/com/vaadin/spring/navigator/SpringNavigator.java#L65)):*
> ```Java
>     private final List<ViewActivationListener> activationListeners = new LinkedList<ViewActivationListener>();
> 
>     public SpringNavigator(ApplicationContext applicationContext, SpringViewProvider viewProvider) {
> ```

adding constructor parameters is API incompatible with 1.1.0

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/157#-:-KX69sG3LHb_GOEnyPKx:bajstg3)*

mstahv:
The fix can be done without field injection "fix" as well. That could be done in a separate change. Also there seems to be another serialisation issue as well.

Will create another PR, closing this.

Agree with @macjuan, there should be, even a very simple integration test for serialisation. Even manual session serialisation would have got this. I'll though let the core team add that change, when ever they have time for it.

PR #168, opened 2016-12-19
Use one shared session destroy listener instead of two per session

### LINKED ISSUE #167: Session cleanup slows down with the number of sessions
Each instance of `UIStore` and `SessionAwareBeanStore` adds itself as a global session destroy listener in the `VaadinService` instance of the corresponding session. This means that every session destroy event is dispatched to all store instances and ignored by all instances except one. This in turn means that session cleanup will get significantly slower as the number of sessions increases.

This should instead be implemented so that there's one shared session destroy listener that inspects the destroyed session to see if it has an associated `UIStore` or `SessionAwareBeanStore` to destroy.

There's also a similar pattern with service destroy listeners, but it's harder to avoid the pattern there since the service destroy event doesn't provide access to all active sessions.

Legioth:
Fixes #167

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/168)


denis-anisimov:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZLhGfsjRh70ZQoazWA:b3fz2oh) (no related file):*
Two public static methods with the same name and same signature.
Doing different things.

May be extract only one method to some util class only one method which does both things ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQJui_iZ9sO0xaVUFE:b-bs2mc) (no related file):*

Two public static methods with the same name and same signature.
Doing different things.

May be extract only one method to some util class only one method which does both things ?
</blockquote></details>

They do the same thing: Clears up the things that the same class has stored in the session. Any difference in how that is done is an implementation detail of each class.

Doing both things in one method would mean that implementation details of each class are handled outside that class.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQOhjI65qlG3bcph7v:b8sqlwh) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

They do the same thing: Clears up the things that the same class has stored in the session. Any difference in how that is done is an implementation detail of each class.

Doing both things in one method would mean that implementation details of each class are handled outside that class.
</blockquote></details>

There something quite wrong with the all design approach used in the `UIScopeImpl` class at least.
And the aspect from here is just a consequence of that.

`UIScopeImpl` can't contain the static method which is introduced here because of the same reason: it's not its implementation detail.
It's an implementation detail of `VaadinSessionBeanStoreRetrievalStrategy` which is just one of the implementation of 
`BeanStoreRetrievalStrategy`. And there  is way to use another one via the `UIScopeImpl. setBeanStoreRetrievalStrategy`.

In this case the code here has no any sense.

It should not be a static method. It should be an instance method and implementation (don't even know impl of what) should implement this method and do it based on its own logic.

But let's not redesign now the whole code in this add-on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQQxJy59xGyFe-YdIa:bmewruu) (no related file):*

There something quite wrong with the all design approach used in the `UIScopeImpl` class at least.
And the aspect from here is just a consequence of that.

`UIScopeImpl` can't contain the static method which is introduced here because of the same reason: it's not its implementation detail.
It's an implementation detail of `VaadinSessionBeanStoreRetrievalStrategy` which is just one of the implementation of 
`BeanStoreRetrievalStrategy`. And there  is way to use another one via the `UIScopeImpl. setBeanStoreRetrievalStrategy`.

In this case the code here has no any sense.

It should not be a static method. It should be an instance method and implementation (don't even know impl of what) should implement this method and do it based on its own logic.

But let's not redesign now the whole code in this add-on.
</blockquote></details>

Actually , I would consider to move the method into `BeanStoreRetrievalStrategy` interface instead.
At least here inside `UIScopeImpl`.

I'm not sure that this is possible because haven't checked runtime details.
But looks like it should be possible.

And something similar to the other static method case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQRN44O1u6Bze26bxn:bbijpoo) (no related file):*

Actually , I would consider to move the method into `BeanStoreRetrievalStrategy` interface instead.
At least here inside `UIScopeImpl`.

I'm not sure that this is possible because haven't checked runtime details.
But looks like it should be possible.

And something similar to the other static method case.
</blockquote></details>

That would mean that the code that stores the attribute in `VaadinSesssion` and the code that uses that attribute to do cleanup would be in different files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQSng7lEhVBr94qhNO:b-1ipoon) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

That would mean that the code that stores the attribute in `VaadinSesssion` and the code that uses that attribute to do cleanup would be in different files.
</blockquote></details>

Sorry , why ?

Currently `UIStore` is an implementation detail of the `VaadinSessionBeanStoreRetrievalStrategy`.

CTOR of `UIStore` set its to the session : 

`java
UIStore(VaadinSession session) {
            sessionId = session.getSession().getId();
            this.session = session;
            this.session.getService().addSessionDestroyListener(this);
            this.session.getService().addServiceDestroyListener(this);
            this.session.setAttribute(UIStore.class, this);
        }
`

And only place where `UIStore`  is used ( if I understand correctly) is `VaadinSessionBeanStoreRetrievalStrategy`.

And it's in the same file as a nested class inside `UIScopeImpl`.

So both has total access to each other.

It's a `BeanStoreRetrievalStrategy` implementation who should know how do a cleanup in reality. It's an implementation detail of this class.
Any other implementation will do something else perhaps ( and it won't know nothing about `UIStore` class of course.

Don't see any issues here.
What am I missing ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZQhc2IaazPmEkL7x39:bx700by) (no related file):*

Sorry , why ?

Currently `UIStore` is an implementation detail of the `VaadinSessionBeanStoreRetrievalStrategy`.

CTOR of `UIStore` set its to the session : 

`java
UIStore(VaadinSession session) {
            sessionId = session.getSession().getId();
            this.session = session;
            this.session.getService().addSessionDestroyListener(this);
            this.session.getService().addServiceDestroyListener(this);
            this.session.setAttribute(UIStore.class, this);
        }
`

And only place where `UIStore`  is used ( if I understand correctly) is `VaadinSessionBeanStoreRetrievalStrategy`.

And it's in the same file as a nested class inside `UIScopeImpl`.

So both has total access to each other.

It's a `BeanStoreRetrievalStrategy` implementation who should know how do a cleanup in reality. It's an implementation detail of this class.
Any other implementation will do something else perhaps ( and it won't know nothing about `UIStore` class of course.

Don't see any issues here.
What am I missing ?
</blockquote></details>

So you're suggesting that the code in the servlet's destroy listener should be `UIScopeImpl.VaadinSessionBeanStoreRetrievalStrategy.cleanupSession(session)` instead of `UIScopeImpl.cleanupSession(session)` and `VaadinSessionScope.VaadinSessionBeanStoreRetrievalStrategy.cleanupSession(session)`  instead of `VaadinSessionScope.cleanupSession(session)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRAJZhIlVvK4EHPzyp:b-dh3a2u) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

So you're suggesting that the code in the servlet's destroy listener should be `UIScopeImpl.VaadinSessionBeanStoreRetrievalStrategy.cleanupSession(session)` instead of `UIScopeImpl.cleanupSession(session)` and `VaadinSessionScope.VaadinSessionBeanStoreRetrievalStrategy.cleanupSession(session)`  instead of `VaadinSessionScope.cleanupSession(session)`?
</blockquote></details>

Yes, as instance methods. And the method should be declared in the `BeanStoreRetrievalStrategy`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRDbB1Y7YSun-eL0Ej:b-o19aza) (no related file):*

Yes, as instance methods. And the method should be declared in the `BeanStoreRetrievalStrategy`.
</blockquote></details>

Adding methods to the `BeanStoreRetrievalStrategy` interface isn't backwards compatible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRK7y4xHbtFcfMZtez:b-lytvkc) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

Adding methods to the `BeanStoreRetrievalStrategy` interface isn't backwards compatible.
</blockquote></details>

OK.

What about different  approach: 
- don't register listener in `SpringVaadinServlet` at all. It really looks as a wrong place for that.
- register a listener in the VaadinSessionBeanStoreRetrievalStrategy (both). They are both singletons in reality. So that there will be only one listener at all per VaadinSessionBeanStoreRetrievalStrategy instance.
- it's even better than listener in servlet since there can be several servlets but "only two" `VaadinSessionBeanStoreRetrievalStrategy`s.
- Now do the same (in any way) but inside implementation classes which know each about other and this will be just an implementation detail of `VaadinSessionBeanStoreRetrievalStrategy`.

Will it work ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRNW9Cd79xYpRJyurl:bu76i3z) (no related file):*

OK.

What about different  approach: 
- don't register listener in `SpringVaadinServlet` at all. It really looks as a wrong place for that.
- register a listener in the VaadinSessionBeanStoreRetrievalStrategy (both). They are both singletons in reality. So that there will be only one listener at all per VaadinSessionBeanStoreRetrievalStrategy instance.
- it's even better than listener in servlet since there can be several servlets but "only two" `VaadinSessionBeanStoreRetrievalStrategy`s.
- Now do the same (in any way) but inside implementation classes which know each about other and this will be just an implementation detail of `VaadinSessionBeanStoreRetrievalStrategy`.

Will it work ?

</blockquote></details>

Each `VaadinSessionBeanStoreRetrievalStrategy` still needs to register the listener for every active instance of `SpringVaadinServlet`. That means that `SpringVaadinServlet.init` would need something like this:
```
UIScopeImpl.VaadinSessionBeanStoreRetrievalStrategy.addSessionDestoryCleanupListener(service);
VaadinSessionScope.VaadinSessionBeanStoreRetrievalStrategy.addSessionDestoryCleanupListener(service);
```

or alternatively, we can change the static `getBeanStoreRetrievalStrategy` methods to return the specific type:

```
UIScopeImpl.getBeanStoreRetrievalStrategy().addSessionDestoryCleanupListener(service);
VaadinSessionScope.getBeanStoreRetrievalStrategy().addSessionDestoryCleanupListener(service);
```

or alternatively, we can make each instance keep track of which `VaadinService` instances it has encountered (i.e. a static ConcurrentHashMap) and register listeners for any instance that still doesn't have the listener. That sounds like a lot of thread synchronization issues just for something that should really happen when the servlets are initialized.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRa-WAXZ8nMJpZ62w2:bno4da8) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

Each `VaadinSessionBeanStoreRetrievalStrategy` still needs to register the listener for every active instance of `SpringVaadinServlet`. That means that `SpringVaadinServlet.init` would need something like this:
```
UIScopeImpl.VaadinSessionBeanStoreRetrievalStrategy.addSessionDestoryCleanupListener(service);
VaadinSessionScope.VaadinSessionBeanStoreRetrievalStrategy.addSessionDestoryCleanupListener(service);
```

or alternatively, we can change the static `getBeanStoreRetrievalStrategy` methods to return the specific type:

```
UIScopeImpl.getBeanStoreRetrievalStrategy().addSessionDestoryCleanupListener(service);
VaadinSessionScope.getBeanStoreRetrievalStrategy().addSessionDestoryCleanupListener(service);
```

or alternatively, we can make each instance keep track of which `VaadinService` instances it has encountered (i.e. a static ConcurrentHashMap) and register listeners for any instance that still doesn't have the listener. That sounds like a lot of thread synchronization issues just for something that should really happen when the servlets are initialized.


</blockquote></details>

Unblocked.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRarsGBw0i_kqdymxj:b-q5jy92) (no related file):*

Unblocked.
</blockquote></details>

Let's see if @peholmst has any ideas on how to handle the unregistration without creating an explicit reference from the servlet to those implementation classes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KZRbehaIjZY2-OVlwvd:b-8fx2u7) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

Let's see if @peholmst has any ideas on how to handle the unregistration without creating an explicit reference from the servlet to those implementation classes
</blockquote></details>

:+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

Legioth:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/spring/168#-KZLhGfsjRh70ZQoazW9:-KcMFzr-7pm5tTrGR39c:bid7saq) (no related file):*

:+1: 
</blockquote></details>

We discussed the matter offline without finding any good solution. It wouldn't be directly inject something or use the Spring event bus since the scope implementations live outside the regular scopes. The alternative would be that we would introduce our own similar abstractions, which would probably be overkill for this case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/168)*

PR #210, opened 2017-04-26
Translate application title (fixes #28)

### LINKED ISSUE #28: Make application title translatable

KyleRogers:
Hi,

what do you think about this fix? Would be glad if you could merge it in. Fixes #28

If you don't like it but know another way to implement this, please elaborate on that approach so that we can consider to implement it and provide a new or updated pull request.

Regards,
Stefan

Note: Replaces #29 and #209
Note 2: Had to change commit author to myself because @benno-mueller-saxess hasn't signed CLA yet. Benno is a collegue of mine and we've implemented the original change together. Thus I've the permission to distribute those lines

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/210)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/annotation/TranslatableTitle.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/annotation/TranslatableTitle.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2015 The original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


pleku:
Hello. Sorry for not responding on this on due time. Thanks for the contribution, but the changes are now unfortunately out-of-date and these are not considered that big improvements that we would want to rebase them and release a new version of v7 version of the add-on at this point.

PR #223, opened 2017-06-27
Let the framework invoke service init listener beans

Legioth:
Integrate discovery of listeners that are spring beans into the regular
init listener discovery so that those listeners will be invoked in the
same way as regular listeners. Without this, additional listeners added
through the session init event would be discarded.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/223)


hesara:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/spring/223)*

PR #242, opened 2017-12-14
Added missing @SpringComponent annotation

wmax:
Because the SpringNavigator seems to expect that the ViewProvider and the ApplicationContext will be injected by springs CDI mechanism.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/242)


tsuoanttila:
Thanks for the PR @wmax, but this is actually not required. There's `VaadinNavigatorConfiguration` that defines a factory method for the `SpringNavigator`.

PR #255, opened 2018-03-14
Fixes erroneous error message when using SpringUIProvider with PushSt…

### LINKED ISSUE #254: Erroneous error message when using SpringUIProvider with PushStateNavigation or wildcard path
SpringUIProvider logs a warning message "Found no Vaadin UIs in the application context" when PushStateNavigation is turned on or when path specifier uses wildcards.

The code seems to be incomplete.
`
if (pathToUIMap.isEmpty()) {   
            logger.warn("Found no Vaadin UIs in the application context");   
}  
`
Maybe should be
       `
 if (pathToUIMap.isEmpty() && wildcardPathToUIMap.isEmpty()) {
            logger.warn("Found no Vaadin UIs in the application context");
        }
`

cyberoblivion:
…ateNavigation or wildcard on the path.

The message "Found no Vaadin UIs in the application context" is logged even when UIs have been found. The current code only checks the pathToUIMap and should check the wildcardPathToUIMap  as well before logging this message.

The message will have a negative impact on users who may think they have incorrectly configured their applications when in fact that is not the case.

closes #254

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/255)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/spring/server/SpringUIProvider.java
+++ b/vaadin-spring/src/main/java/com/vaadin/spring/server/SpringUIProvider.java
@@ -85,7 +85,7 @@ public class SpringUIProvider extends UIProvider {
                     "Spring WebApplicationContext not initialized for UI provider. Use e.g. ContextLoaderListener to initialize it.");
         }
         detectUIs();
-        if (pathToUIMap.isEmpty()) {
+        if (pathToUIMap.isEmpty() && wildcardPathToUIMap.isEmpty()) {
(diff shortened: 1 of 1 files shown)


tsuoanttila:
Hi @cyberoblivion. Sorry to be so late at responding to your PR. There has been a massive refactor in the repository which makes the current version of the PR target the wrong branch. Can you open another PR targeting `3.0` instead of `master`?

cyberoblivion:
NP, Will do

PR #287, opened 2018-04-19
Avoid singleton component in test class

Legioth:
Use of a globally shared component instance will be broken after vaadin/flow#3797

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/287)

### DIFF
--- a/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/FooNavigationTarget.java
+++ b/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/FooNavigationTarget.java
@@ -15,12 +15,11 @@
  */
 package com.vaadin.flow.spring.test;
 
-import org.springframework.beans.factory.config.ConfigurableBeanFactory;
-import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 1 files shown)


denis-anisimov:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/FooNavigationTarget.java, line 30 at r1](https://reviewable.io/reviews/vaadin/spring/287#-LDaYElsAvD_m7I3akGN:-LDaYElt6VbxYxss6p0h:b-w8m510) ([raw file](https://github.com/vaadin/spring/blob/5545dfb4612577a2dca8a5727fbf6bf9be9d457a/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/FooNavigationTarget.java#L30)):*
> ```
> @VaadinSessionScope
> ```

Does it fix the issue in #308.
I expect the same session which means that the instance will be reused on refresh.
Which means that there will be the same exception.

PR #314 uses `UIscope` and changes the logic of IT test so that the instance won't be preserved in refresh anymore (but IT test should follow the router link to pass).

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/spring/287)*

Issue #292, opened 2018-04-23
Memory leak: VaadinSession and session scoped beans not cleaned up due to bug in UIScopeImpl


hmtelbi:
Vaadin Spring Version: 3.0

I noticed that VaadinSessions, VaadinSessionScope$SessionAwareBeanStore and all session scoped beans are not garbage collected.
It seems that the root cause can be found in the destroy method of UIScopeImpl, where the session attribute, which contains the SessionAwareBeanStore (Key: BeanStore.class) is set to null instead of the attribute for the UIStore (Key: UIStore.class):

```
        void destroy() {
            LOGGER.trace("Destroying [{}]", this);
            session.setAttribute(BeanStore.class, null);
            session.getService().removeServiceDestroyListener(this);
            for (BeanStore beanStore : new HashSet<BeanStore>(
                    beanStoreMap.values())) {
                beanStore.destroy();
            }
            Assert.isTrue(beanStoreMap.isEmpty(),
                    "beanStore should have been emptied by the destruction callbacks");
        }
```

As a consequence, the beanStore is always null in the cleanupSession method of VaadinSessionScope and thus the bean store is never destroyed:

```
    public static void cleanupSession(VaadinSession session) {
        assert session != null;

        BeanStore beanStore = session.getAttribute(BeanStore.class);
        if (beanStore != null) {
            LOGGER.debug("Vaadin session has been destroyed, destroying [{}]",
                    beanStore);

            beanStore.destroy();
        }
    }
```

PR #332, opened 2018-08-06
Fix memory leak: Session BeanStore.class was unset from Session

herby3:
Session BeanStore.class was unset from Session before Session clean up (see #292)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/332)


herby3:
I created a test project here: https://github.com/herby3/VaadinSessionCleanUp
The @PreDestroy cleanUp() method of ExampleSessionBean is never called after closing the vaadin session, because the BeanStore.class was unset to null too early. So the Session-Beans are no longer known. With my patch the ExampleSessionBean is destroyed and the @PreDestroy method is called.

mehdi-vaadin:
Fixes #292

PR #335, opened 2018-08-13
added hotfix for spring boot 2.0.4 compatibility

mstahv:
Added a quick workaround for 2.0.4 compatibility. With next maintenance release notes we'll need to mention that 2.0.4 is required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
 
         <!-- These are typically overridden with BOMs -->
         <vaadin.flow.version>1.1-SNAPSHOT</vaadin.flow.version>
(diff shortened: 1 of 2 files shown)


Legioth:
Was this an unexpected breaking change in Spring, or is there something we could have done to avoid this problem altogether?

mstahv:
Yes and yes. Spring folks expected that nobody is using that and we are probably using that in a wrong way. That's a long story that we could have tomorrow at the office with @snicoll, @pleku and you.

mstahv:
Nope, that's why I added the comment to PR.

PR #336, opened 2018-08-13
Extract new DispatcherServletRegistrationBean into its own config

denis-anisimov:


vaadin-spring/src/main/java/com/vaadin/flow/spring/DispatcherServletRegistrationBeanConfig.java
```diff
@@ -0,0 +1,65 @@
...
+ * This is a workaround for spring boot 2.0.4 compatibility (see spring#331).
+ *
+ * @see #dispatcherServletRegistration()
+ *
+ * @author Vaadin Ltd
+ *
+ */
+@Configuration
+@Conditional(RootMappedCondition.class)
+@ConditionalOnClass(name = "org.springframework.boot.autoconfigure.web.servlet.DispatcherServletRegistrationBean")
```
snicoll: For the record, there is no need to use a name here, you can use the FQN. We use ASM to apply those conditions without loading the class.

This works when `@ConditionalOnClass` is at class level only (as in this case).

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -93,20 +94,22 @@ public ServletContextInitializer contextInitializer() {
...
      *
-     * @return a custom DispatcherServletRegistrationBean instance for dispatcher servlet
+     * @return a custom DispatcherServletRegistrationBean instance for
+     *         dispatcher servlet
      */
     @Bean
     @Conditional(RootMappedCondition.class)
-    public DispatcherServletRegistrationBean dispatcherServletRegistration() {
+    @ConditionalOnMissingClass("org.springframework.boot.autoconfigure.web.servlet.DispatcherServletRegistrationBean")
+    public ServletRegistrationBean<DispatcherServlet> dispatcherServletRegistration() {
```
snicoll: As there is a extra condition, it would be worth adding a note as why it is necessary?

PR #343, opened 2018-08-28
Use Spring Boot autoconfigure-processor to optimize auto-configurations

snicoll:
Closes gh-328

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/343)


pleku:
Thanks for the contribution Stéphane! 

I'm not sure whether this change is backwards compatible to existing V10 projects - if they update their projects to a vaadin-spring version that includes this change, there might be some automatic configuration changes caused by this change ? Is there possibility that those override any manual configurations ?

Or is this just to make it faster to start the Spring boot ? (Sorry but I'm not sure of all the implications)

PR #363, opened 2018-10-15
Fixes #362

### LINKED ISSUE #362: Cannot specify an interface as a layout in @Route
If you have the following interface:
```Java
interface Layout extends RouterLayout {
}
```
And a corresponding implementation:
```Java
@SpringComponent
@Scope("prototype") // or session or UI scope alternatively
public class LayoutImpl extends Composite<VerticalLayout> implements Layout {
    ...
}
```
You cannot specify the interface as a layout in a Router view as follows:
```Java
@Route(layout = Layout.class) // note Layout is used, not LayoutImpl
public class View extends Composite<VerticalLayout> {
    ...
}
```
An example of when this would be useful is in multi-module applications where an API is offered to allow extending the functionality of the app. An API module containing the interface can be distributed without having to distribute the implementation.

In order to fix this, Flow's internal `AnnotationReader` class needs the fix made in [this commit](https://github.com/vaadin/flow/commit/399424cef1a8989bb05cc9d6e589b3801e196b26#diff-fdc5711112ad890b94c7f315dc1b7297).

alejandro-du:
This tries to get a bean from the application context before attempting to create a new instance of the bean.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/363)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
@@ -19,6 +19,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
 
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
 
 import com.vaadin.flow.di.DefaultInstantiator;
@@ -81,10 +82,11 @@ public class SpringInstantiator extends DefaultInstantiator {
 
     @Override
     public <T> T getOrCreate(Class<T> type) {
-        if (context.getBeanNamesForType(type).length == 1) {
+        try {
             return context.getBean(type);
-        }
 
-        return context.getAutowireCapableBeanFactory().createBean(type);
+        } catch (NoSuchBeanDefinitionException e) {
+            return context.getAutowireCapableBeanFactory().createBean(type);
+        }
     }
 }


alejandro-du:
___
*[vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java, line 88 at r1](https://reviewable.io/reviews/vaadin/spring/363#-LOrgEnaDKsQHR3bOVOR:-LOrio4C-OXIEHhNQiFy:b28aa17) ([raw file](https://github.com/vaadin/spring/blob/a1b463f2a46f5274b60bfefa8d0791a0542bc6f7/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java#L88)):*

> ```
> } catch (NoSuchBeanDefinitionException e) {
> ```

We should avoid catching unchecked exception whenever it's possible.

Doesn't this work : `context.getBeanNamesForType(type).length == 1)` ?
The whole idea of this code was to avoid any exception and check whether there are beans with the required type....
</blockquote></details>

It doesn't work when there is​ more than one implementation of the bean, say, one marked with `@Primary`. But the following should work as well, if you prefer not catching the exception: `context.getBeanNamesForType(type).length > 0`

alejandro-du:
___
*[vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java, line 88 at r1](https://reviewable.io/reviews/vaadin/spring/363#-LOrgEnaDKsQHR3bOVOR:-LOsUsQy5DaptQj9SBxi:brxkgd9) ([raw file](https://github.com/vaadin/spring/blob/a1b463f2a46f5274b60bfefa8d0791a0542bc6f7/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java#L88)):*

`context.getBeanNamesForType(type).length > 0` is wrong since if there are many eligible beans then `context.getBean(type)` thrown an exception.
</blockquote></details>

Not if you mark one as `@Primary`.

alejandro-du:
___
*[vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java, line 88 at r1](https://reviewable.io/reviews/vaadin/spring/363#-LOrgEnaDKsQHR3bOVOR:-LOw5q4W2mMKRilp8gE6:b-xirb5o) ([raw file](https://github.com/vaadin/spring/blob/a1b463f2a46f5274b60bfefa8d0791a0542bc6f7/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java#L88)):*

And there should be a test for this.
</blockquote></details>

Throwing an exception in this situation is what I would expect in a Spring app. The same would happen if you try to inject a bean with two possible implementations that are not somehow qualified (e.g. no @Primary annotation) in any Spring app. But there's also another problem I see with creating a new instance: It might not follow the scope a developer intends (e.g. singleton or session).

pleku:
@alejandro-du please respond to the requested changes or just close this PR. Or we will soon close it too because it is no point in leaving it hanging here.

PR #408, opened 2019-03-15
Fix for "Misleading Error Message when autowiring beans" in Issue #407

JulianFeinauer:
When using SpringBeans with Vaadin for autowiring, there is a misleading error message when one has 2 or more suitable Beans in the Context.
The SpringInstantiator's getOrCreate Method autowires the bean only when there is exactly one bean in the scope or otherwise it tries to instantiate the given interface, which leads to the message

`Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [this.is.my.Interface]: Specified class is an interface`

This PR fixes this and shows a separate exception when there are more than 1 beans in the context, and does not try to instantiate a new bean instance.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/408)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java
@@ -83,8 +83,11 @@ public class SpringInstantiator extends DefaultInstantiator {
     public <T> T getOrCreate(Class<T> type) {
         if (context.getBeanNamesForType(type).length == 1) {
             return context.getBean(type);
+        } else if (context.getBeanNamesForType(type).length > 1) {
+            throw new IllegalStateException("Unable to autowire existing beans, there are more " +
+                    "than 1 autowiring candidates");
         }
-
+        // If there is no bean, try to instantiate one
         return context.getAutowireCapableBeanFactory().createBean(type);
     }
 }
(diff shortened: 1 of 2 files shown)


JulianFeinauer:
One question for the Reviewer: I do not like to throw the `IllegalStateException` there. But I did not see a good example what Kind of exception you prefer there, so please give me some advice there.

JulianFeinauer:
@denis-anisimov thanks for your review. I see that there are some more issues on how the spring integration should actually work and I agree.
The main issue I had (#407) is that it always told me that it cannot instanciate an interface although I had beans (but sadly I had two). Thus this is more of a minor fix to make it a step better (my colleague just had the same exception and spend one hour investigating) than a refactoring (which I agree... is totally necessary.
I do not like your idea so much to make this based on abstract class and interface. If I have a (non-abstract) base-class and a subclass instance as bean this should still autowire, cause I expect it to autowire. Otherwise I would always have to check what kind of superclass a given instance has.

I will definetly add a test, if you agree with my fix. All i wanted to do is to report a proper Failure Message when it cannot autowire a bean due to having multiple ones.

JulianFeinauer:
@denis-anisimov I added the fix.
I suggest that we open another discussion for the contract of the getOrCreate Method, what do you think?

PR #412, opened 2019-03-25
Prevents @Id from injecting unrelated beans

AdrienC:
See https://github.com/vaadin/flow/issues/4634

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/412)


denis-anisimov:
Thank you for your contribution but this is not a proper fix since it introduces a regression.

PR #498, opened 2019-09-25
Adapt Spring add-on classes search logic for full CP scanning

### LINKED ISSUE #497: Make sure that fix for flow#6480 works with Spring add-on
Two things:
- Spring uses its own way to scan classes. So the current add-on state needs adaptation to be able to work with full classpath scanner: the annotation set which is scanned should be changed.
- During the investigation ticket (for full CP scanning) it has been discovered that `JsModule` annotations (looks like any annotation) is not found on superclass like `Button` (which is OK since it's not inheritable and its super class has it so the superclass should be found) and it's not found for `GeneratedVaadinButton` (this is not OK). `GeneratedVaadinButton`  is completely ignored: it's not discovered anyhow.

So:
- fix the first item. That's easy. Only change the annotations collection.
- check the second item: whether all required classes are discovered ? Add tests for this: no need to check full CP scanner. It's already properly tested in its own PR. But `ClassFinder` which is used in Spring add-on should be somehow tested.

denis-anisimov:
Fixes #497

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/498)

### DIFF
--- a/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/UploadIT.java
+++ b/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/UploadIT.java
@@ -19,6 +19,7 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.List;
 
 import org.junit.Assert;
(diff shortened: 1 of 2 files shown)


denis-anisimov:
This PR requires https://github.com/vaadin/flow/pull/6545

Issue #519, opened 2019-12-02
JsModules from interfaces are not added to frontend bundle


johannesh2:
- Using Spring starter with 14.0 and updating it to 14.1.
- With a simple view and components implementing DragSource and DropTarget
- Compile and run in development mode
The resulting application throws client-side exceptions for `cannot read property 'updateDragSource' of undefined` and `cannot read property 'updateDropTarget' of undefined`.
It seems that dndConnector.js is not included

pleku:
Can be reproduced with the following code:

```
@Route
@PWA(name = "Project Base for Vaadin Flow with Spring", shortName = "Project Base")
public class MainView extends VerticalLayout {

    public MainView(@Autowired MessageBean bean) {
        Button button = new Button("Click me",
                e -> Notification.show(bean.getMessage()));
        DragSource.create(button);

        Button anotherButton = new AnotherButton();
        anotherButton.setText("Another");
        DragSource.create(anotherButton);

        Div div = new Div();
        div.setWidth("300px");
        div.setHeight("50px");
        div.getStyle().set("background", "pink");
        DropTarget.create(div).setDropEffect(DropEffect.MOVE);

        add(button, anotherButton, div);
    }

    public class AnotherButton extends Button implements DragSource<Div> {
    }
}

```

The `dndConnector.js` is not loaded into the browser (or the es6 one). 

Can we workaround with using `@@JsModule(DndUtil.DND_CONNECTOR)` in the `MainLayout`, and the it works.

PR #525, opened 2019-12-05
Scan abstract types

Legioth:
Reverts the hardcoded workaround for #519 and instead configures the
classpath scanning logic to consider all abstract types.

This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/525)


caalador:
`SpringClassesSerializableTest` fails

Issue #478, opened 2019-08-26
[Vaadin Spring] RootMappedCondition not completely working


mixaaaa:
Hi,

since spring boot 2 the relaxed property binding behavior had been changed and does not work like `RootMappedCondition.class` uses it at the moment. Because of that the usage of property `vaadin.url-mapping` does not work instead of `vaadin.MappingUrl` which perfectly works :)

The following example shows the usage of new property binding api which should work

```java
	@Override
	public boolean matches(ConditionContext context,
			AnnotatedTypeMetadata metadata) {
		VaadinConfigurationProperties vaadinConfigurationProperties = Binder.get(context.getEnvironment())
				.bind("vaadin", VaadinConfigurationProperties.class)
				.orElse(null);
		if (vaadinConfigurationProperties != null) {
			return isRootMapping(vaadinConfigurationProperties.getUrlMapping());
		}
		return false;
	}

	/**
	 * Returns {@code true} if {@code mapping} is the root mapping
	 * ({@literal "/*"}).
	 * <p>
	 * The mapping is controlled via the {@code vaadin.urlMapping} property
	 * value. By default it's {@literal "/*"}.
	 *
	 * @param mapping
	 *            the mapping string to check
	 * @return {@code true} if {@code mapping} is the root mapping and
	 *         {@code false} otherwise
	 */
	public static boolean isRootMapping(String mapping) {
		if (mapping == null) {
			return true;
		}
		return mapping.trim().replaceAll("(/\\**)?$", "").isEmpty();
	}
```

Cheers & Thanks for your feedback,
Uwe

mehdi-vaadin:
@mixaaaa sorry for such a late reply. It seems that you are proposing a change in `RootMappedCondition` here. Would you like to create a PR for it?

PR #537, opened 2020-01-21
[spring#478] Bugfix

mixaaaa:
+ switched to springs binding api to fix root mapped condition for vaadin flow

Example: https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#migrating-from-1x-to-20
Link: https://github.com/vaadin/spring/issues/478

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/537)


knoobie:
@denis-anisimov How about using `Class::forName` to check if `org.springframework.boot.context.properties.bind.Binder` is present? So spring boot users can benefit from the feature and the other users get the current behaviour?

knoobie:
@denis-anisimov It's not broken, it's just not as feature rich as the spring boot equivalent. You can find more about it in the [docs](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-relaxed-binding) or [wiki](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0) - the relaxed binding is a spring boot only feature.

mixaaaa:
Hey :)

Sorry for late response but i do not really get why we should not use Spring Boot feature in a spring boot condition. 

Would it not also be possible to move the static method into a util class  for accessing it from `VaadinMVCWebAppInitializer` as same as from Spring Boot Condition class `RootMappedCondition`? So in future for Spring Boot Users with auto configuration it is perfectly working and users directly using MVCInitializr must use the exact property. 

If this is no option than we could try initializing a conditional bean on spring boot reading properties if possible and inject it into our `RootMappedCondition`. By the way i have to check if this is an option. I'm not pretty sure.

Cheers & stay healthy, Uwe

mixaaaa:
Check, got your point. Sorry, wasn't much time too look for it yesterday.

I added some changes to search for our configuration properties in the bean factory. My problem is i can not run the build at the moment because I'm missing vaadin snapshot dependencies.

Could you test this solution do anyone see any problem about this change? :)

Cheers, Uwe

pleku:
The build is still failing apparently to a regression (?) introduced by this change @mixaaaa 
https://teamcity.vaadin.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=66494&_focus=1528

pleku:
Closing this PR due to inactivity for a long time and as the team has other priorities that go above fixing this.

In case this is still important, please reopen this on top of the latest changes from master and fix the tests.

PR #539, opened 2020-01-23
Add pnpm.enable as a config parameter name for Spring

### LINKED ISSUE #538: Flag reverting pnpm is broken
For 14.2, when writing the following to `application.properties`, it should switch between `pnpm` and `npm`:
```
vaadin.disable.pnpm=true
```
But this not happen, project is always run with `pnpm`

NOTE: this will be changed to `vaadin.pnpm.enable` after vaadin/flow#7398

denis-anisimov:
* Add pnpm.enable config property name
* Rename test so that it became a test

Fixes #538

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/539)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
@@ -15,14 +15,14 @@
  */
 package com.vaadin.flow.spring;
 
+import javax.servlet.ServletException;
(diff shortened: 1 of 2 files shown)


denis-anisimov:
The build is failing because:
- it uses `npm` in dev mode instead of pnpm. That's an issue in Flow
- another question why `npm`  takes _so much time_ that spring boot is stopped by timeout. Because `npm`  is absolutely valid tool which can be used here.

PR #563, opened 2020-02-28
update flow and project versions

haijian-vaadin:
- Update Flow version to 3.1-SNAPSHOT
- Update project version to 13.1-SNAPSHOT
- Add 2 new constants REQUIRE_HOME_NODE_EXECUTABLE and SERVLET_PARAMETER_MAX_MESSAGE_SUSPEND_TIMEOUT for the test to make the validation pass. Since the new constant was introduced in  Flow 3.1-SNAPSHOT

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/563)


haijian-vaadin:
Also need to fix the test when updating the Flow version, because the new constant was introduced in 3.1-SNAPSHOT. Otherwise, the validation of the PR fails

PR #594, opened 2020-04-05
Cherry pick from 12.2 due to skinnywar fix in Flow 3.0

haijian-vaadin:
- Use process method for ServletContextInitializers instead of onStartup (#591)
- Use pnpm to build IT modules

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/594)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -293,15 +294,15 @@ public void failFastContextInitialized(ServletContextEvent event) {
 
         @SuppressWarnings("unchecked")
         private void validateAnnotations(
-                ServletContainerInitializer initializer, ServletContext context,
-                List<Class<?>> annotations) {
+                ClassLoaderAwareServletContainerInitializer initializer,
```
pleku, line 297: @denis-anisimov is this the only place that needs changing to add the new `ClassLoaderAwareServletInitializer` ?

There are a lot of forked initializers in use in this class since 3.0 and it uses its own `VaadinServletContextInitializer` implementation so I'm doubtful that this change based on a cherry-pick is enough 

```
public class VaadinServletContextInitializer
        implements ServletContextInitializer {	        implements ServletContextInitializer {
```

PR #595, opened 2020-04-06
Use process method for ServletContextInitializers instead of onStartup (#591)

denis-anisimov:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/595)

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -357,8 +357,10 @@ public class VaadinServletContextInitializer
                 throw new RuntimeException(
                         "Unable to initialize Vaadin DevModeHandler", e);
             }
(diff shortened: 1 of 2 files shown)


mshabarov:
> Repeat my comment again: there is no need to catch this and rethrow a runtime exception.

Yes, I agree. There is no extra information in Runtime exception thrown by subclasses and anyway `ServletException` is caught on `FailFastServletContextListener` level in default method and properly handles it. So, I would exclude those catches from `DevModeServletContextListener`, `WebComponentServletContextListener` and `AnnotationValidatorServletContextListener`

denis-anisimov:
> So, I would exclude those catches from ....

?????

The comment is attached to the code which is changed in this PR.
Before this PR there was no any `catch`.  Now it's there.
There is no reason why it's there.
It should be reverted.
Other listeners are not touched in this PR. 
If changes are needed there they can be done in another PR.
This PR needs to be fixed.

mshabarov:
:lgtm:

PR #600, opened 2020-04-08
Fix enabling pnpm from application.properties

### LINKED ISSUE #597: Regression for reading vaadin.pnpm.enable in 14.2
Since 12.2.0.alpha4 (14.2.0.alpha10, Flow 2.2.0.alpha16) the `vaadin.pnpm.enable=true` property in `application.properties` is not working for switching to pnpm (as now npm is again the default).
It doesn't matter whether coming from npm 14.1 or 14.2 (using the alpha10). 
It doesn't matter whether running via Maven `spring-boot:run` or from `Application` class.
It doesn't matter if the project is cleaned or not (no pre-existing tokenfile).

In Vaadin 15.0.3, the configuration works and pnpm is enabled.

When forced from Maven plugin configuration
```
<configuration>
    <pnpmEnable>true</pnpmEnable>
</configuration>
```
pnpm works as it is written in the token file correctly `"pnpm.enable": true,`

When using `14.2.0.alpha9`, switching from default pnpm to npm using the same property `vaadin.pnpm.enable=false` does work and npm is used instead.

denis-anisimov:
Fixes #597 

This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/600)

### DIFF
--- a/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
+++ b/vaadin-spring-tests/test-spring-boot-only-prepare/pom.xml
@@ -108,9 +108,6 @@
                 <version>${vaadin.flow.version}</version>
                 <executions>
                     <execution>
-                        <configuration>
-                            <pnpmEnable>true</pnpmEnable>
-                        </configuration>
                         <goals>
                             <goal>prepare-frontend</goal>
                         </goals>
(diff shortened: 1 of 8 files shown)


mshabarov:
:lgtm:

PR #608, opened 2020-04-25
add bean validation implementation dependency

### LINKED ISSUE #606: vaadin-spring-boot-starter 15.0.5 projects do not start with Spring Boot 2.3
Spring Boot 2.3 has removed the validation implementation `spring-boot-starter-validation` from `spring-boot-starter-web`, see https://github.com/spring-projects/spring-boot/issues/19550. If you take the Spring starter today and upgrade to 2.3.0.M3, startup fails with
```
***************************
APPLICATION FAILED TO START
***************************

Description:

The Bean Validation API is on the classpath but no implementation could be found

Action:

Add an implementation, such as Hibernate Validator, to the classpath
```

haijian-vaadin:
Fixes #606

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/608)

### DIFF
--- a/vaadin-spring/pom.xml
+++ b/vaadin-spring/pom.xml
@@ -82,6 +82,13 @@
 
         <!-- End Spring -->
 
+        <!-- Bean validation implementation-->
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <version>6.1.4.Final</version>
+        </dependency>
+
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>


Artur-:
I think you should add `spring-boot-starter-validation` instead so that the version comes from Spring Boot

snicoll:
> I think you should add spring-boot-starter-validation instead so that the version comes from Spring Boot

It is unusual for a library to bring a starter in compile scope and the starter is not a requirement at all to get dependency management. If the version in the current dependency is removed and this project is configured to use Spring Boot's dependency management, the version of `hibernate-validator` will be indeed driven by the version of Spring Boot the project compiles against.

haijian-vaadin:
Hi @Artur-, I think probably using hibernate is better. Since  `spring-boot-starter-validation` includes some other dependencies that we don't really need:
1. `jakarta.validation-api`, we already have the `javax.validation.validation-api` in Flow
2. `tomcat-embed-el`, which we don't really use/need currently.

Also using `spring-boot-starter-validation` would force the user to use Spring boot, even though (theoretically) we should support non-boot Spring applications.

What do you think?

PR #615, opened 2020-05-02
Flow Spring 12.2 nightly build for Bower mode fails #605

### LINKED ISSUE #605: Flow Spring 12.2 nightly build for Bower mode fails
https://bender.vaadin.com/viewLog.html?buildId=141258&tab=buildResultsDiv&buildTypeId=Flow_AddOns_FlowSpring122BowerMode

mshabarov:
Apply another approach to fix Spring 12.2 builds failures in Bower mode: disable fork mode for spring-boot-maven-plugin to avoid issues with JMX.

Fixes #605

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/615)


mshabarov:
*[a discussion](https://reviewable.io/reviews/vaadin/spring/615#-M6TmRVF1hJJtTzNlkDg:-M6ToD_RC3nPRvtm_dbq:b-naauxj) (no related file):*

There are quite many changes.
Are they all related to the fix ?
At least commit message says only about fork for plugin but in fact there are source code changes.
</blockquote></details>

Yes, they all related to the fix. There is basically three parts: 1. Delete previous code with ports, it didn't improve the issue with a build 2. Disable fork mode. 3 Add Vaadin parameters explicitly via System.setProperty() since the setting of them in pom.xml is ignored in fork disabled mode.

---

mshabarov:
___
*[vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/util/TestUtils.java, line 38 at r1](https://reviewable.io/reviews/vaadin/spring/615#-M6YILqG11mHOrLkY3e8:-M6_H73u7HhE37ELWMVz:bsv7hp0) ([raw file](https://github.com/vaadin/spring/blob/3ab54742f84445c7054ffa8569409d7010dc4157/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/util/TestUtils.java#L38)):*

> ```
> properties.setProperty("vaadin.compatibilityMode", "");
> ```

property can be (and should be) specified in `application.properties`  file.
Any reason to do it this way instead of `app.properties`?
Also it should be possible somehow to set an env property via Spring API at the application startup.
This code may go to the `common` module and invoked via `@Configuration` for every module.

This way looks ugly for me (just because system properties). I would go with it in the worst case.
</blockquote></details>

Yes, agreed. I decided to avoid application.properties for last time, however I can't recall why at the moment. So, yes, they should be there.

mshabarov:
___
*[vaadin-spring-tests/test-spring-boot/src/main/java/com/vaadin/flow/spring/test/TestServletInitializer.java, line 39 at r1](https://reviewable.io/reviews/vaadin/spring/615#-M6YHfzW3Ajj6rmpPCOO:-M6_HtpgEQom4HxUBvwk:b-skc0f1) ([raw file](https://github.com/vaadin/spring/blob/3ab54742f84445c7054ffa8569409d7010dc4157/vaadin-spring-tests/test-spring-boot/src/main/java/com/vaadin/flow/spring/test/TestServletInitializer.java#L39)):*

> ```
>         TestUtils.setVaadinProperties();
>         SpringApplication.run(TestServletInitializer.class, args);
> ```

This is repeating code.
I would extract a class into `common` module and pass parameters to a method (maybe a static one) : `Class<?>, String[]`.
To avoid duplicating code.

Or may be reuse `TestUtils` for this method.
</blockquote></details>

TestUtils has been removed and properties added to app.properties instead.

caalador:
The addition of `vaadin.compatibilityMode=` to application properties makes the application run in the wrong mode making `Vaadin Spring Boot integration tests with custom packages to scan` fail with no es6 packages found. Also I guess other modules now run in compatibility not npm mode.

mshabarov:
> The addition of vaadin.compatibilityMode= to application properties makes the application run in the wrong mode

Yes, the behaviour become even worse after switching fork mode off. I believe we have to make a step back and find another solution for the source problem - floating issue with JMX/RMI.

PR #626, opened 2020-05-21
Flow Spring 12.2 nightly build for Bower mode fails #605

### LINKED ISSUE #605: Flow Spring 12.2 nightly build for Bower mode fails

mshabarov:
Fixes #605 

This PR is another attempt to workaround the bower mode build issue by disabling the fork mode for spring-boot-maven-plugin. Last time there was a silly mistake - compatibility mode has been set for all the tests which made the application run in wrong mode and led to test failures.

Now, the compatibility mode is set exactly when the bower mode (bower profile) is detected. The implementation is still using System.setProperty(), since it is critical to set those properties before Spring Application starts in order to pass the properties to Vaadin services. Spring application startup callback API (CommandLineRunner, ApplicationEvent) are invoked too late for this and can't be used.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/spring/626)

### DIFF
--- a/vaadin-spring-tests/pom.xml
+++ b/vaadin-spring-tests/pom.xml
@@ -28,13 +28,6 @@
         <driver.binary.downloader.maven.plugin.version>1.0.14
         </driver.binary.downloader.maven.plugin.version>
         <maven.war.plugin.version>3.1.0</maven.war.plugin.version>
-
-
-        <!-- Specify the same port for both JMX and RMI in order to avoid the issues
-             with JMX connection between spring-boot-maven-plugin and Spring Boot application
-             Workaround for https://github.com/vaadin/spring/issues/605 -->
-        <spring.jmxremote.port>9001</spring.jmxremote.port>
-        <spring.rmi.port>9001</spring.rmi.port>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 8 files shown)


knoobie:
Just wondering: why can't you add `vaadin.compatibilityMode=true` to the application.properties? Or when it has to be dynamic like this `vaadin.compatibilityMode=${bowerMode}`?

mshabarov:
> Fork is disabled: it means that all processes are executed in the same JVM, right ?
> It means that once vaadin.compatibilityMode is set it never reset to false.
> Is it right ?

Yes, that's right. However, the build is started (i.e. JVM started) by Bender agent with a bower profile. That means that this `vaadin.compatibilityMode` property will be there within that JVM process which is limited only to 'Vaadin Spring Boot integration tests ... in bower mode' step. So other steps in Bender (other JVMs) would not be affected by this property.
So I don't see a difference between `vaadin.compatibilityMode` property set by `spring-boot-maven-plugin` as it's being done currently and this System.setProperty approach.

Where am I wrong?

mshabarov:
> Edit 2: Just an idea to remove the additional system properties you are going to create - how about manipulating the String[] args and append the property there? Like you can see here Command line arguments have a higher precedence than Java System properties as well.

Alright, I will check that. Thanks!

mshabarov:
@denis-anisimov , @knoobie I've tried to add to command-line args array, but Flow reads this property from a system properties or init parameters, so it is invisible for Vaadin application and it fails to start.

knoobie:
@mshabarov thanks for looking into this! 

Totally out of scope, but sounds like a good addition for me. Related code for future generations:

https://github.com/vaadin/spring/blob/f554a3d5d0e6f109d8cc5e27b8dee1bd859f4381/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java#L132-L138

Issue #634, opened 2020-06-09
Spring boot configuration vaadin.pnpm-enabled


jcgueriaud1:
In Vaadin 14.3.0.alpha1, when I tried to update Spring configuration file to switch to PNPM, by default IntelliJ suggests this:
vaadin.pnpm-enabled=true/false

Which is not working.

The documentation says:
For a Spring Boot based project, you can put vaadin.pnpm.enable = true into the application.properties files

Which is the right value, any idea why this configuration is found?

caalador:
the property is `pnpm.enable` so it means in the application.properties it should be `vaadin.pnpm.enable`

jcgueriaud1:
In this file  https://github.com/vaadin/spring/blob/master/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
 the property is pnpmEnabled. That's exactly what IntelliJ found and suggest in the autocomplete.

jcgueriaud1:
I think the problem is related to this PR: https://github.com/vaadin/spring/pull/572

pleku:
@denis-anisimov what do you think, should we just revert the contribution that adds it for autoconfiguration since it suggests the wrong format ?

denis-anisimov:
Yes, let's remove it.

I can't find how to use `ConfigurationProperties` with properties which have dots in the name.

snicoll:
> I can't find how to use ConfigurationProperties with properties which have dots in the name.

What do you mean by that? 

> , should we just revert the contribution that adds it for autoconfiguration since it suggests the wrong format ?

What wrong format is that? What is the difference between `asyncSupported` and `pnpmEnabled` in the very same class?

This should provide the following properties:

* `vaadin.async-supported` (default to `true`)
* `vaadin.pnpm-enabled` (default to `false`)

I don't really understand the problem. There is a typo in the description of the `pnpmEnabled` property.

jcgueriaud1:
The correct property is vaadin.pnpm.enable.

The file VaadinConfigurationProperties.java contains vaadin.pnpm-enabled and has been made to provide IDE assistance. This property is not used.

The other properties in this java file are used (I think).

Perhaps there is a different way to provide IDE assistance like manually update or create META-INF/spring-configuration-metadata.json for all the vaadin configuration properties.
vaadin.pnpm.enable
vaadin.productionMode ....

That will be really useful.

snicoll:
[...]

PR #635, opened 2020-06-10
Remove the pnpmEnabled property and add documentation 

jcgueriaud1:
Add documentation for vaadin properties that are not bound to a @ConfigurationProperties bean to provide IDE Assistance.

Here is the result before the changes.
* vaadin.compatibility is not found and IntelliJ adds a warning
* no IDE Assistance

vaadin_properties_before

After the changes:
* IntelliJ does not complain anymore
* IDE Assistance: default value and documentation

after_spring_changes

Related to this ticket: https://github.com/vaadin/spring/issues/634

I didn't add all the value documented here: https://vaadin.com/docs/v14/flow/advanced/tutorial-all-vaadin-properties.html

I'm not sure if they should all be here (For example i18n.provider is not required in a Spring Application)


snicoll (commented):
I wonder if some of those properties couldn't be in the main code. In particular heartbeatInterval.

vaadin-spring/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,40 @@
+{
+  "properties": [
+    {
+      "name": "vaadin.productionMode",
```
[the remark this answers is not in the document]
jcgueriaud1: I'm not sure if I understand the suggestion.
The parameter should be according to the spring format: `vaadin.production-mode`

Currently it's `vaadin.productionMode` because the property name comes from vaadin flow. It's the same name in a Vaadin project without Spring prefixed by vaadin.

Do you mean that if I change the metadata to `"name":` `"vaadin.production-mode"` the existing application will still work with `vaadin.productionMode` ?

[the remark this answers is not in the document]
snicoll: The canonical format is in no way an "enforced" format but the format you must use so that your metadata is compilant.

> Do you mean that if I change the metadata to "name": "vaadin.production-mode" the existing application will still work with vaadin.productionMode ?

Yes.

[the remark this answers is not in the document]
jcgueriaud1: I've just created a ticket about it, thanks to your comment.
https://github.com/vaadin/spring/issues/637

vaadin-spring/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,40 @@
...
+      "defaultValue": false
+    },
+    {
+      "name": "vaadin.pnpm.enable",
+      "type": "java.lang.Boolean",
+      "description": "This flag can be used to enable pnpm instead of npm for resolving and downloading frontend dependencies.",
+      "defaultValue": false
+    }
+    ]
+}
```
denis-anisimov, line 40: new line

denis-anisimov (changes requested):
The list of properties is not complete but I'm OK with that.
The problem with the resource file : it's static and has no any checks against the real properties.

We won't be able to note that something has changed in Flow (some property is removed/added/changed) to update this file.
So I'm pretty sure once written it won't be updated ever.

About removing the `pnpmEnabled` : it's OK here, but I think we should keep it in the correct way.
I will create another PR which corrects the property name/usage.

jcgueriaud1:
The flow properties are read from application.properties in the  SpringServlet#getInitParameterNames()
So you shouldn't read these properties a second time in VaadinConfigurationProperties.java. VaadinConfigurationProperties should be only for spring-specific properties (like blacklisted-packages...).

If you check the spring documentation here: https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html#configuration-metadata-additional-metadata
It's written:
`Spring Boot’s configuration file handling is quite flexible, and it is often the case that properties may exist that are not bound to a @ConfigurationProperties bean. You may also need to tune some attributes of an existing key. To support such cases and let you provide custom "hints", the annotation processor automatically merges items from META-INF/additional-spring-configuration-metadata.json into the main metadata file.`

That's, in my opinion, the case for all flow configuration properties, including vaadin.pnpm.enable.

And unfortunately you can't generate this, it's a double maintenance (if it's written as Java or json). Since Spring can't generate automatically these metadata.

denis-anisimov:
>VaadinConfigurationProperties should be only for spring-specific properties

No, this doesn't fit to the users expectation.
`pnpmEnabled` property has been added by the external contributor.
We have an issue about inability to use `VaadinConfigurationProperties` because they don't set env variable (we read only env properties).

So Spring boot users prefer to use typesafe way to define the properties  instead of `application.propeprties` file.
Unfortunately `VaadinConfigurationProperties`  doesn't work at the moment at all since it is not reflected to the env properties as I expected but this is another issue.

Yes, we can't generate it and that's a serious issue which makes it almost useless.

But I don't insist on avoiding this way. Let it be.

jcgueriaud1:
I'm not sure if I understand this:
> No, this doesn't fit to the users expectation.

The user expectations in this PR https://github.com/vaadin/spring/pull/572 is to permit IDE autocomplete. Or is there something else?

(Vaadin)ConfigurationProperties is to access to properties defined in properties files (and spring auto-generate the metadata that is used in the IDE).

miguelatvaadin:
My plan so far:

- write a unit test which checks if `additional-spring-configuration-metadata.json` is synchronized with `VaadinConfigurationProperties.java`
- add/remove whatever is needed to `additional-spring-configuration-metadata.json` from the docs

jcgueriaud1:
- write a unit test which checks if additional-spring-configuration-metadata.json is synchronized with VaadinConfigurationProperties.java

You don't need to add the properties defined in VaadinConfigurationProperties in an additional-spring-configuration-metadata.json. It's done automatically by Spring.

- add/remove whatever is needed to additional-spring-configuration-metadata.json from the docs

Yes, that would help the developer experience.

miguelatvaadin:
> You don't need to add the properties defined in VaadinConfigurationProperties in an additional-spring-configuration-metadata.json. It's done automatically by Spring.

Better. Less work to do ;) Thanks for the comment, @jcgueriaud1

miguelatvaadin:
Ok. After a deeper review of spring's documentation and some quick tests I will change my plan to:

- write a unit test to verify that all Vaadin init parameters are in `VaadinConfigurationProperties`
- write a unit test to verify that the properties are used when initializing the servlet
- add init parameters to `VaadinConfigurationProperties`, so they are recognized by the IDE
- get rid of `additional-spring-configuration-metadata.json`

miguelatvaadin:
After implementing the last solution proposal I have realized that the IDE, by default, proposes kebab-case naming for the configuration property names. That means that instead of `vaadin.productionMode` the ide proposes `vaadin.production-mode`, even though both alternatives (camel and kebab) are still valid due to Spring's relaxed parsing.

It's a little weird as properties in doc are stated as camel case, so I will do some research to figure out if we can set camel case as the default for our configuration properties.

After that little research I will clear everything up and push the changes for reviewal.

jcgueriaud1:
> It's a little weird as properties in doc are stated as camel case, so I will do some research to figure out if we can set camel case as the default for our configuration properties.

In my opinion, a spring configuration should follow the Spring rule because this is expected by a Spring developer.

I would prefer that Flow follows the same rule but it doesn't:
https://github.com/vaadin/flow/issues/9420

A note in the documentation could help if needed.

miguelatvaadin:
Thanks for your comments, @jcgueriaud1 

> In my opinion, a spring configuration should follow the Spring rule because this is expected by a Spring developer.

Yes, that's my opinion too, indeed. Just thinking it would have been more natural for me to find the same names as in the doc.

> I would prefer that Flow follows the same rule but it doesn't

By moving the parameters to `VaadinConfigurationProperties`  it will be followed, indeed.

> A note in the documentation could help if needed.

Yes. Absolutely. Surely that will be the way to go.

miguelatvaadin:
I have just pushed the code changes.

At the end I have not been able to write any unit test as my code, indeed, does nothing. It's there just to let spring generate the appropiate autocompletion hints for the IDE. In fact, the ' VaadinConfigurationProperties' object is not used to read the parameters but they are read from the environment instead.

By adding the parameters to `VaadinConfigurationProperties` I think everything behaves as @jcgueriaud1 expected, and spring's kebab-case properties can be used.

Validation of parameters happens later when creating the deployment configuration object, so I have not added any bean validation annotation to the newly created properties.

miguelatvaadin:
After a nice talk with @jcgueriaud1 , I will add some code for solving an little problem related to the `vaadin.urlMapping` property, so the `VaadinConfigurationProperties` class is fully operational.

miguelatvaadin:
`vaadin.urlMapping` is solved and a unit test added for all the new properties from `VaadinConfigurationProperties`

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+     * relaxed parsing
+     *
+     * @param env the environment
+     * @param propertyName the relaxed property name
+     * @return the actual property name
+     */
+    public static String getActualPropertyName(Environment env, String propertyName) {
+        String storedPropertyName = null;
+        String normalizedPropertyName = normalizePropertyName(propertyName);
+        Iterator it = ((AbstractEnvironment) env).getPropertySources().iterator();
```
denis-anisimov, line 46: How it's known that the `Environment` instance is always an `AbstractEnvironment` subclass ?
Several questions here:
- is it always true ?
- Shouldn't be there some check whether it's a subclass at least using assertion or even throw a meaningful exception ?

denis-anisimov: Another question : Iterator is generic.
Here is has no type. Is it so that the type is not known and varies ? I would consider this is as a weird API in Spring.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+        while(it.hasNext() && storedPropertyName == null) {
+            PropertySource propertySource = (PropertySource) it.next();
```
denis-anisimov, line 48: Shouldn't be this a type in the `Iterator` from the very beginning ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+            if (propertySource instanceof MapPropertySource) {
+                storedPropertyName = extractFromPropertySource(
+                        propertySource, normalizedPropertyName);
```
denis-anisimov, line 51: This call should have a case to `MapPropertySource` right away so that the method `exrtactFromPropertySource` does something meaningful only for this type.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+        return storedPropertyName;
+    }
+
+    private static String extractFromPropertySource(
+            PropertySource propertySource, String normalizedPropertyName) {
```
denis-anisimov, line 58: `propertySource` argument should have `MapPropertySource` type right away since it's suggested in the code.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+    private SpringHelper() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    /**
+     * Gets the property name used after being parsed by Spring Boot, accepting
+     * relaxed parsing
+     *
+     * @param env the environment
+     * @param propertyName the relaxed property name
```
denis-anisimov, line 40: can the parameter be `null` ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return the actual property name
```
denis-anisimov, line 41: Can it return a `null` ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+        return storedPropertyName;
+    }
+
+    private static String extractFromPropertySource(
+            PropertySource propertySource, String normalizedPropertyName) {
+        String storedPropertyName = null;
+        for (String key : ((MapPropertySource) propertySource)
+                .getSource().keySet()) {
+            if (normalizePropertyName(key).equals(normalizedPropertyName)) {
```
denis-anisimov, line 62: Normalization process may results in unexpected matches.

E.g. `foo__bar` will match `foo_bar`.
Is it OK ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
```diff
@@ -161,6 +161,11 @@ private Properties config(Properties initParameters) {
         properties.putAll(initParameters);
         PROPERTY_NAMES.forEach(property -> setProperty(property, properties));
 
+        // this property is reversed
+        if ("true".equalsIgnoreCase(
```
denis-anisimov, line 165: `true`  is `Boolean.TRUE` in a typesafe way.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
```diff
@@ -161,6 +161,11 @@ private Properties config(Properties initParameters) {
...
+                initParameters.getProperty("disable-xsrf-protection")))
```
denis-anisimov, line 166: What is `disable-xsrf-protection` ?
Is it a known constant ?  Then it should be used as a constant here.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
```diff
@@ -161,6 +161,11 @@ private Properties config(Properties initParameters) {
...
+            properties.put("xsrf-protection-enabled", false);
```
denis-anisimov, line 167: What does this code do ?

`InitParameters.SERVLET_PARAMETER_DISABLE_XSRF_PROTECTION` is an init parameter which is used in the deployment configuration.

`initParameters` are servlet init parameters.
Produced `properties` are used for deployment configuration . 
Deployment configuration doesn't know anything about `xsrf-protection-enabled`. 
So I don't understand why this code exists.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
```diff
@@ -87,6 +88,152 @@ public void setEnable(boolean enable) {
...
 
+    /**
+     * Turns application to work in production mode. Production mode disables
+     * most of the logged information that appears on the console because
+     * logging and other debug features can have a significant performance
+     * impact. Development mode JavaScript functions are not exported, push is
+     * given as a minified JavaScript file instead of full size and static
+     * resources are cached. See Deploying to Production for more information.
+     */
+    private Boolean productionMode;
```
denis-anisimov, line 99: Is this class gives the IDE assistance ( instead of `additional-spring-configuration-metadata.json` ) ?

vaadin-spring/src/test/java/com/vaadin/flow/spring/SpringHelperTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SpringHelperTest {
+
+    @Test
+    public void camelCase_normalized_matches() {
+        String normalized = SpringHelper
+                .normalizePropertyName("vaadin.urlMapping");
+        Assert.assertEquals("vaadinurlmapping", normalized);
```
denis-anisimov, line 40: This is exactly my concern. 
Now I can use `vaadinurlmapping` for `vaadin.url-mapping` .
So : 
- `vaadin` prefix is not a prefix anymore
- the `urlmapping` is just the same as  `url-mapping`.

vaadin-spring-tests/test-spring-boot/src/main/resources/application.properties
```diff
@@ -1 +1 @@
-server.port=8888
+server.port=8888
```
denis-anisimov, line 1: extra change: should be reverted

denis-anisimov (changes requested):

vaadin-spring-tests/test-spring-boot/src/main/resources/application.properties
miguelatvaadin: ok. reverted

vaadin-spring/src/main/java/com/vaadin/flow/spring/RootMappedCondition.java
```diff
@@ -43,8 +44,13 @@
...
             AnnotatedTypeMetadata metadata) {
-        return isRootMapping(
-                context.getEnvironment().getProperty(URL_MAPPING_PROPERTY));
+        String mapping = getUrlMappingParameter(context.getEnvironment());
+        return isRootMapping(mapping);
+    }
+
+    private String getUrlMappingParameter(Environment environment) {
+        String propertyName = SpringHelper.getActualPropertyName(environment, URL_MAPPING_PROPERTY);
+        return propertyName != null?environment.getProperty(propertyName):null;
```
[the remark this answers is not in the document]
miguelatvaadin: ok., Thanks

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.PropertySource;
+
+/**
+ * Utility methods for spring
+ */
+public class SpringHelper {
+
+    private SpringHelper() {
+        throw new IllegalStateException("Utility class");
```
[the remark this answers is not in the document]
miguelatvaadin: ok. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: ok. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: ok. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Gets the property name used after being parsed by Spring Boot, accepting
+     * relaxed parsing
+     *
+     * @param env the environment
+     * @param propertyName the relaxed property name
+     * @return the actual property name
+     */
+    public static String getActualPropertyName(Environment env, String propertyName) {
```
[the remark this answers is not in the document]
miguelatvaadin: buf! changed to `Stored`

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: done. I'm afraid I copy-pasted this code without further revision

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: done. the same as above

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: yes. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
miguelatvaadin: for me it is a valid result. I just want to check that they go to the same place. Also, the method is only checking vaadin properties, so I think it is acceptable to think that no conflicts will happen

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
miguelatvaadin: ok. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
miguelatvaadin: done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
miguelatvaadin: I do not know, either. I remove it.

vaadin-spring/src/test/java/com/vaadin/flow/spring/SpringHelperTest.java
miguelatvaadin: As commented before, I think this is correct. Well, I would better say that it works / is acceptable.

I would prefer to use the read the values directly from `VaadinConfigurationProperties`, so all this code would be unnecessary. In fact, this is what I did at the begining, until I discovered the MVC part of the integration and I realized that, yes or yes, I needed to rely on properties.

Once that we accept that we must rely on properties, and that spring accepts that relaxed naming, we must be able to find the property whatever name it is given. Hence all this code.

Anyway, as said before, I think it is feasible to say that we will not use `vaadinurlmapping` and `vaadin.url-mapping` for different properties.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,87 @@
...
+                                               String propertyName) {
+        assert env != null;
+        assert propertyName != null;
+
+        String storedPropertyName = null;
+        String normalizedPropertyName = normalizePropertyName(propertyName);
+        if (env instanceof AbstractEnvironment) {
+            for (PropertySource<?> propertySource :
+                    ((AbstractEnvironment) env).getPropertySources()) {
+                if (propertySource instanceof MapPropertySource) {
```
denis-anisimov, line 53: `if` is inside `for` inside `if`.

It's too hard to read this code: please refactor and extract something to a method.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Gets the property name used after being parsed by Spring Boot, accepting
+     * relaxed parsing
+     *
+     * @param env the environment, not {@code null}
+     * @param propertyName the relaxed property name, not {@code null}
+     * @return the stored property name. {@code null} if not found
+     */
+    public static String getStoredPropertyName(Environment env,
```
denis-anisimov, line 43: If method may return `null` then it's better to return `Optional`: it makes the contract clearer.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
denis-anisimov: I didn't get why the `Environment` instance is always a `AbstractEnvironment` .
There is class cast without any checks and any explanation.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringHelper.java
denis-anisimov: This becomes kind of internal API.
So anyone who updates Spring code may use this method .
The contract doesn't limit anyone from using the method only for Vaadin properties which are hardcoded.

The result is not valid for me generally.
It has a lot of issues :
- `foo.bar` property is normalized as `foobar` which means the property `bar` in the `foo` namespace is the same as `foobar` which is incorrect
- property `foo__bar` is the same as `foobar`.
- property `foo--bar` is the same as `foobar`.

Neither _existing constant_ Vaadin property has `_` /`-` inside the name.
So it's not a Vaadin property but rather something which _may_ become a Vaadin property. And the examples  shown above may convert a property which should not be a Vaadin property into Vaadin property.

The last concern: I cannot find the usages of `SpringHelper` class except the only one inside `RootMappedCondition`.
Am I missing anything ?
I don't see any point to have such method for only one usecase which can be handled properly for only one property.

vaadin-spring/src/test/java/com/vaadin/flow/spring/SpringHelperTest.java
denis-anisimov: It doesn't matter whether we use `vaadinurlmapping` or not: this is not a Vaadin property at all.
The property  `vaadinurlmapping`  may be accidentally used by the developer and it should not match `vaadin.url-mapping`.

Again: the current set of properties and the code are not final. So it means that the code is evolving.
At some point the same code from `SpringHelper` can be used for some other property and then the mistake won't be so obvious.

I don't see :
- reasons to have this as a utility method
- it looks like it's used only in one specific case and it's easier to make normalization/converstation directly as a private method specifically for this case.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
denis-anisimov: This comment is not resolved.

And in fact I don't like the way how this done.
I still don't get whether this gives out of the box the IDE assistance .

Another problem again: this is hardcoded and doesn't allow to code evolve.
Every time when property name is changed in Flow or a new one is added this needs to be changed and no-one 
will remember to do it.

So such approach either requires a test which checks that all the properties are used here 
OR
the approach which I prefer is: _generate_ a file based on the Vaadin properties.

Both files can be generated : Java and json.
Then the question how to include them into the project: the project maven config needs to be changed to include them.
Perhaps it's easier to generate json file and include a resource into the jar.
But it's up to you.

miguelatvaadin: Well, the truth is that I understand and agree with your concerns, Denis.

They lead me to a complete redesign of the PR, though, so I will think a little about it and post a new design to the issue as soon as I feel it's good enough.

Then just write the code and push the PR again.

miguelatvaadin:
This is the new design for the PR, which supports the following additional requisites as extracted from Denis' comments:

- generate the spring boot configuration class so it always remains synchronized with the accepted Vaadin servlet properties
- allow exactly (no more than) the relaxed property naming scheme of spring boot

### Description

We will generate the `VaadinConfigurationProperties` class using an annotation processor, which will act on a new class where we will put any property which we want to have into our final `VaadinConfigurationProperties` class (I mean the properties which are used by the spring initializers). The annotation processor will also add all of the servlet properties which are present in the `InitParameters` class.

So, the current propeties from `VaadinConfigurationProperties` will be moved to a new class called `VaadinConfigurationPropertiesTrigger` which will contain those properties which are used by the spring initializers. This class will be also be annotated with a newly created annotation called `VaadinConfigurationPropertiesGenerator`, which will trigger the annotation processor so the final `VaadinConfigurationProperties` will be created. 

We will use a @PostConstruct method on the generated VaadinConfigurationProperties class to dump all the read properties to the enviroment/system properties with the original expected property name so the servlet/Spring initializers can use them, without "touching" anything else.

The annotation processor will be in a new module inside the project, as it must be compiled before the spring project itself, and the dependency  will be scoped for compile time.

pleku:
>This is the new design for the PR, which supports the following additional requisites as extracted from Denis' comments:
> - generate the spring boot configuration class so it always remains synchronized with the accepted Vaadin servlet properties
> - allow exactly (no more than) the relaxed property naming scheme of spring boot

Looks good to me 👍🏻

caalador:
Seems good.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -82,11 +82,16 @@ public ServletContextInitializer contextInitializer() {
...
-                .setAsyncSupported(configurationProperties.isAsyncSupported());
+                .setAsyncSupported(
+                        configurationProperties.getAsyncSupported() == null ?
+                                true : // default value
+                                configurationProperties.getAsyncSupported());
         registration.setName(
                 ClassUtils.getShortNameAsProperty(SpringServlet.class));
         registration.setLoadOnStartup(
-                configurationProperties.isLoadOnStartup() ? 1 : -1);
+                configurationProperties.getLoadOnStartup() == null ||
```
denis-anisimov, line 92: Same question here: should the return type be `boolean` and not `Boolean`
And then the getter name can be preserved as `isXXX`

vaadin-spring/src/main/resources/META-INF/services/javax.annotation.processing.Processor
```diff
@@ -0,0 +1 @@
+com.vaadin.flow.spring.ConfigurationPropertiesAnnotationProcessor
```
denis-anisimov, line 1: new line

vaadin-spring/src/test/java/com/vaadin/flow/spring/configuration/VaadinConfigurationPropertiesCamelCaseTest.java
```diff
@@ -0,0 +1,53 @@
...
+        "vaadin.maxMessageSuspendTimeout=600",
+        "vaadin.pushUrl=AAA",
+        "vaadin.requestTiming=true",
+        "vaadin.syncIdCheck=true",
+        "vaadin.useDeprecatedV14Bootstrapping=true",
+        "vaadin.sendUrlsAsParameters=true"
+}, classes = { VaadinConfigurationProperties.class })
+public class VaadinConfigurationPropertiesCamelCaseTest extends VaadinConfigurationPropertiesTest {
+
+}
```
denis-anisimov, line 53: new line

vaadin-spring/src/test/java/com/vaadin/flow/spring/configuration/VaadinConfigurationPropertiesNegativeTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    public void hasConfiguration_xsrfProtectionEnabled_servletSet() throws ServletException {
+        VaadinService service = SpringInstantiatorTest.getService(context,
+                new Properties());
+        boolean xsrfProtectionEnabled = service.getDeploymentConfiguration()
+                .isXsrfProtectionEnabled();
+        Assert.assertEquals(false, xsrfProtectionEnabled);
+    }
+}
```
denis-anisimov, line 81: new line

vaadin-spring/src/test/java/com/vaadin/flow/spring/configuration/VaadinConfigurationPropertiesTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Test
+    public void hasConfiguration_pushURL_servletSet() throws ServletException {
+        VaadinService service = SpringInstantiatorTest.getService(context,
+                new Properties());
+        String pushURL = service.getDeploymentConfiguration().getPushURL();
+        Assert.assertEquals("AAA", pushURL);
+    }
+
+}
```
denis-anisimov, line 179: new line

denis-anisimov (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -82,11 +82,16 @@ public ServletContextInitializer contextInitializer() {
                 new SpringServlet(context, rootMapping), mapping);
         registration.setInitParameters(initParameters);
         registration
-                .setAsyncSupported(configurationProperties.isAsyncSupported());
+                .setAsyncSupported(
+                        configurationProperties.getAsyncSupported() == null ?
```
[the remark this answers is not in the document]
miguelatvaadin: the configuration properties are all created as objects to know if the value comes from the application.properties or not (null values case). In this case the value is indeed initialized and that check should not be necessary but, as it can happen, I added the check. 

I could indeed check if there is a default value for the property and, if this happens, use the primitive instead of the object version, but I didn't want to write a more complex code for this. If you still think that something must be done, I could go for this

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
miguelatvaadin: more or less the same as before. I can implement a more complex code generation logic if you think it's worthed

vaadin-spring/src/main/resources/META-INF/services/javax.annotation.processing.Processor
miguelatvaadin: mmm... it seems I lost my style configuration. I'll fix all the new lines you commented. Thanks

miguelatvaadin:
I have added a new module and moved som classes to avoid bringing javaparser and flow source as transitive dependencies to projects using the spring addon

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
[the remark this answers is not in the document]
denis-anisimov: >the configuration properties are all created as objects to know if the value comes from the application.properties or not

Does it matter whether the property is defined in the `application.properties` (which is in fact not the only way of setting the properties) or not ?

Previous  version of the class was also based on properties (let's say defined in `application.properties` ) but it had a default value `false` it the property is not available.

I understand that the code generation will be more complicated if we take into account boxing types (primitives and their boxed Java types).
And I'm kind if OK to avoid spending time on this considering that _most of_ properties are not used in fact via `VaadinConfigurationProperties`.

The only thing which I care:
- this PR changes _existing_ API. 
- previously there was `isAsyncSupported` with `boolean` as a property 
- previously there was `isLoadOnStartup` with `boolean` as a property 

Now the API is changed and the only reason why: the automatic class generation whose purpose is just provide IDE assistance.

I don't think it's a good thing to change API with a reason which has no relation to the API itself but rather just because of simplicity to generate the code.

So:
- I don't insist on _generic_ handling primitives in a proper way
- but API for `isLoadOnStartup` and `isAsyncSupported` should not be changed.

It's up to you how you want to deal with this: 
- consider these two properties as a special case and somehow "hardcode" them
- or fix them generally for any type.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
denis-anisimov: The point is : API is changed just because of code generation simplicity.

See above.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+@SupportedAnnotationTypes("com.vaadin.flow.spring.SpringConfigurationPropertiesGenerator")
+public class ConfigurationPropertiesAnnotationProcessor extends AbstractProcessor
+{
+
+    private static final String INITPARAMETERS_CLASS_NAME = "com.vaadin.flow.server.InitParameters";
+    private Messager messager;
+    private Map<String, CompilationUnit> compilationUnitCache = new HashMap<>();
+
+    @Override
+    public synchronized void init(ProcessingEnvironment processingEnv) {
```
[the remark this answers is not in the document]
miguelperezcolom: the method is already defined as `synchronized` in the superclass

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        super.init(processingEnv);
+        messager = processingEnv.getMessager();
+        messager.printMessage(Diagnostic.Kind.WARNING, "hola!");
```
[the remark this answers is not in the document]
miguelperezcolom: Ups! removed

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    class FieldDescriptor {
```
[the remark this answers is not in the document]
miguelperezcolom: sure. Thanks!

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        final String fieldName;
```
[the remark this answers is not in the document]
miguelperezcolom: I intentionally left them as package protected to avoid declaring getters, as it is an internal use class which is used for holding some info only

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        public FieldDescriptor(String formerPropertyName, String type,
+                               String defaultValueSetter, String comments) {
+            assert formerPropertyName != null;
+            assert type != null;
+            this.formerPropertyName = formerPropertyName;
+            this.defaultValueSetter = defaultValueSetter;
+            this.fieldName = toCamelCase(formerPropertyName, false);
+            // we do not want primitives, as we want any property to be nullable
+            // to know if the value has actually been set by the user
+            if ("int".equals(type)) type = "Integer";
```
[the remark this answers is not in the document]
miguelperezcolom: done. This piece of code has been moved to another place, also.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+    public boolean process(Set<? extends TypeElement> annotations,
+                           RoundEnvironment roundEnv) {
+        messager.printMessage(Diagnostic.Kind.NOTE,
+                "Starting annotation processing from " + getClass().getName());
+        annotations.forEach(annotation -> processAnnotation(roundEnv, annotation));
+        return false;
+    }
+
+    private void processAnnotation(RoundEnvironment roundEnv, TypeElement annotation) {
+        roundEnv.getElementsAnnotatedWith(annotation).forEach(e -> {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. Done. Thanks!

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+
+            // SpringConfigurationPropertiesGenerator is restricted to Types, so it's safe to cast the element
+            addFieldsFromTypeElement((TypeElement) e, fields);
+
+            addFieldsFromInitParametersClass(fields);
+
+            // same package, VaadinConfigurationProperties as class name
+            String annotatedClassName = e.toString();
+            String generatedClassName = e.getEnclosingElement().toString()
+                    + "." + "VaadinConfigurationProperties";
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+
+            String prefix = getPrefix((TypeElement) e);
+
+            try {
+                writeJavaCodeToFile(annotatedClassName, generatedClassName,
+                        prefix, fields);
+            } catch (IOException ioException) {
+                messager.printMessage(Diagnostic.Kind.ERROR, ioException.getClass().getSimpleName() + ": " + ioException.getMessage());
```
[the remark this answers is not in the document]
miguelperezcolom: "Printing a message with an error kind will raise an error"

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                e.getAnnotation(
+                        SpringConfigurationPropertiesGenerator.class);
+        String prefix = annotationInstance.prefix();
+        return prefix;
+    }
+
+    private void addFieldsFromInitParametersClass(List<FieldDescriptor> fields) {
+        String pathToInitParameters = "/" + INITPARAMETERS_CLASS_NAME
+                .replaceAll("\\.", "/") + ".java";
+        InputStream r = getClass().getResourceAsStream(pathToInitParameters);
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        if (parsed.getResult().isPresent()) {
+            addFieldsFromCompilationUnit(parsed.getResult().get(), fields);
+        } else {
+            messager.printMessage(Diagnostic.Kind.ERROR,
+                    pathToInitParameters
+                            + " is missing. Please check if the flow-server sources are available as a dependency");
+        }
+    }
+
+    private void addFieldsFromCompilationUnit(CompilationUnit cu,
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                                              List<FieldDescriptor> fields) {
+        cu.accept(new VoidVisitorAdapter<List<FieldDescriptor>>() {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+            @Override
+            public void visit(FieldDeclaration fieldDeclaration,
+                              List<FieldDescriptor> arg) {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+
+                if (hasInitialValue(fieldDeclaration)) {
+                    FieldDescriptor fd = new FieldDescriptor(
+                            fieldDeclaration.getVariable(0).
+                                    getInitializer().get().
+                                    toStringLiteralExpr().get().
+                                    getValue(),
+                            fieldDeclaration.getElementType().toString(),
+                            null, null);
+                    fieldDeclaration.getComment().ifPresent(c ->
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                }
+            }
+        }, fields);
+    }
+
+
+    private void addFieldsFromTypeElement(TypeElement initParams,
+                                          List<FieldDescriptor> fields) {
+        processingEnv.getElementUtils().getAllMembers(initParams).stream()
+                .filter(m -> ElementKind.FIELD.equals(m.getKind()))
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                .filter(m -> m instanceof VariableElement) // in theory all fields are VariableElements
+                .map(m -> (VariableElement) m)
+                .map(m -> new FieldDescriptor(m.getSimpleName().toString(),
+                        m.asType().toString(),
+                        getInitialValueAssignmentCode(m),
+                        processingEnv.getElementUtils().getDocComment(m)))
+                .forEach(fields::add);
+    }
+
+    private String getInitialValueAssignmentCode(VariableElement m) {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        String className = m.getEnclosingElement().toString();
+
+        CompilationUnit cu = getCompilationUnit(className);
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        String value = getInitialValueAssignmentCode(cu, m.getSimpleName().toString());
+
+        return value;
+    }
+
+    private CompilationUnit getCompilationUnit(String className) {
+        return compilationUnitCache.computeIfAbsent(className, key -> {
+            String pkgName = key.substring(0, key.lastIndexOf('.'));
+            String simpleClassName = key.substring(key.lastIndexOf('.') + 1);
+            CompilationUnit cu = null;
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                    cu = parsed.getResult().get();
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            return cu;
+        });
+    }
+
+    private String getInitialValueAssignmentCode(CompilationUnit cu, String fieldName) {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        assert fieldName != null;
+
+            String[] value = new String[1];
+            if (cu != null) {
+                cu.accept(new VoidVisitorAdapter<String[]>() {
```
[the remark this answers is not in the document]
miguelperezcolom: Just following the visitor pattern as expected. Not sure how to avoid using some value holder (e.g `String[]`) by introducing a nested class here.

Anyway, I will change the `arg` name

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+            }, value);
+        }
+        return value.toString();
+    }
+
+    private void writePostConstruct(PrintWriter out,
+                                    List<FieldDescriptor> fields) {
+        out.println(" @PostConstruct");
+        out.println(" public void transferValuesToSystem() {");
+        fields.forEach(f -> {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

denis-anisimov:
Nothing is done in the PR or not committed (which is the same)

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+            out.println("  if (" + f.fieldName + " != null) " +
+                    "System.setProperty(\"vaadin." + f.formerPropertyName
```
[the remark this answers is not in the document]
miguelperezcolom: In fact the generated configuration class itself is not used anywhere but in `SpringBootAutoConfiguration`, and only for the `urlMapping` parameter. All of the other values inside the configuration must be sent to the vaadin servlet somehow. 

The vaadin servlet reads these values via the `DeplymentConfiguration` class, which reads the values from web.xml or from the system properties. Hence setting system.properties.

Anyway, I chose this option as I understood it was less intrusive (less original code modified). The other option I managed was to serialize all the values to a Map, so it could be passed to the servlet as the initial parameters inside the `SpringBootAutoConfiguration::servletRegistrationBean` method.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+
+    private void writeField(PrintWriter out, FieldDescriptor field) {
+        out.println();
+
+        // write comment
+        String comment = field.comments;
+        if (comment != null && !"".equals(comment)) {
+            out.println("  /**");
+            Arrays.stream(comment.split("\n")).
+                    forEach(l -> out.println("  * " + l));
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+            out.println("  */");
+        }
+        // write field
+        out.print("  private " + field.type + " " + field.fieldName);
+        if (field.defaultValueSetter != null)
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        out.println("  public void set" + capitalize(field.fieldName)
+                + "(" + field.type + " " + field.fieldName + ") {");
+        out.println("    this." + field.fieldName + " = " +
+                field.fieldName + ";");
+        out.println("  };");
+
+    }
+
+    private String toCamelCase(String value, boolean firstUppercase) {
+        if (value == null) return null;
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        StringBuffer camelized = new StringBuffer();
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        Arrays.stream(value
+                .replaceAll("([A-Z])", "\\-$1")
+                .split("[\\.\\-_)]"))
```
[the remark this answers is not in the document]
miguelperezcolom: well, the generated configuration class plains all of the property names. Why? because we have properties like `pnpm.enable` or `devmode.live-reload.enabled` which use dots though they are not intended to be complex objects. 

In fact, all of those values are read as simple String values from the vaadin servlet, so I understood it made no sense to generate a complex classes structure for those cases.

All this also enables using `pnpm-enable` as much as `pnpm.enable`, which was something expected.

Also, `toCamelCase` is intended for generating a method name, so dots must be converted to something to generate a valid method name.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                .forEach(t -> camelized.append(capitalize(t)));
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        return firstUppercase ? Utils.capitalize(camelized.toString()) :
+                Utils.decapitalize(camelized.toString());
+    }
+
+    private String capitalize(String t) {
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        assert t != null;
+        return t.isEmpty() ? "" : Utils.capitalize(t);
+    }
+
+    private String cleanComment(String comment) {
+        /*
+        comments from t compilation unit come with initial asterisks and spaces
```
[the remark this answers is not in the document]
miguelperezcolom: Ups! it should say `the`. Thanks

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        so we want to remove them and have the clean comment text
+        */
+        StringBuffer clean = new StringBuffer();
```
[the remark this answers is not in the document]
miguelperezcolom: ok. Changed to StringBuilder

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        if (comment != null) {
+            Arrays.stream(comment.split("\n")).
+                    filter(l -> !"".equals(comment.trim()))
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                    .forEach(l ->
+                            clean.append(
+                                    ("".equals(clean.toString()) ? "" : "\n")
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+                                            + l.replaceAll("^[\\* ]+",
+                                            "")));
+        }
+        return clean.toString();
+    }
+
+}
```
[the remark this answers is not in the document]
miguelperezcolom: ok. done

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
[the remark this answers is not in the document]
miguelperezcolom: ok. The annotation processor has been generalized so `isLoadOnStartup` and `isAsyncSupported` remain

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
miguelperezcolom: same as above. Though I see this code is missing, now.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
denis-anisimov: So why do you need to duplicate it here ?

`synchronized` in the super class doesn't mean that you have to mark the overriden method .
`synchronized` in the superclass protects superclass state for some reasons. And it will protect it regardless of whether you  repeat it here or not : the super method is called .

So I didn't get the answer.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
denis-anisimov: I don't see how package protected visibility change anything:
- if the class and its fields are used outside of the class (inside other class in the package) then accessor methods are absolutely required.
- if they are used only inside the class then `private` doesn't prevent to use them inside the class. Every field must be `private` .

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -300,38 +283,62 @@ private void writePostConstruct(PrintWriter out,
                                     List<FieldDescriptor> fields) {
         out.println(" @PostConstruct");
         out.println(" public void transferValuesToSystem() {");
-        fields.forEach(f -> {
-            out.println("  if (" + f.fieldName + " != null) " +
-                    "System.setProperty(\"vaadin." + f.formerPropertyName
-                    + "\", " + f.fieldName + ".toString());");
+        fields.forEach(field -> {
+            String code = "  if (" + field.fieldName + " != null) " +
+                    "System.setProperty(\"vaadin." + field.formerPropertyName
```
denis-anisimov, line 288: `System.setProperty` : what does this code do ?
Why does it set System property ?

Do you set every property from `application.properties` as a system ?
What's the reason for this ?

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -300,38 +283,62 @@ private void writePostConstruct(PrintWriter out,
...
+            if (isPrimitive(field)) {
+                code = "  System.setProperty(\"vaadin." + field.formerPropertyName
+                        + "\", \"\" + " + field.fieldName + ");";
+            }
+            out.println(code);
         });
         out.println(" }");
     }
 
+    private boolean isPrimitive(FieldDescriptor f) {
```
denis-anisimov, line 299: one letter variable

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
```diff
@@ -300,38 +283,62 @@ private void writePostConstruct(PrintWriter out,
...
+        boolean primitive = boolean.class.toString().equals(f.type)
```
denis-anisimov, line 300: There are more primitive types.
`Class.isPrimitive` does what you need.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
denis-anisimov: We are talking about totally different things.

There is no sense to generate complex structures: yes.

You are saying that "pnpm.enable" property in `application.properties` is the same as `pnpm-enable`.
This is UNEXPECTED to me (since I thought that Spring at least uses prefix `"vaadin."  for a namespace  which means that `"vaadin.foo"` can't be the same as `"vaadin-foo"` because they are in different namespaces.
May be I'm wrong and they are treated in the same way by Spring: I would like to see the confirmation of this.

denis-anisimov (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
```diff
@@ -100,4 +98,13 @@ public ServerEndpointExporter websocketEndpointDeployer() {
         return new VaadinWebsocketEndpointExporter();
     }
 
+    /**
+     * Makes the request util available.
```
denis-anisimov, line 102: Why ?
For what?
Where this class is used ?

denis-anisimov (changes requested):

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
denis-anisimov: Not done.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
denis-anisimov: This is totally different issue.

And it's already exists: https://github.com/vaadin/spring/issues/478 (or almost the same).

This is another aspect of Spring: it doesn't map `Environment` properties and properties class but I would expected this:
- _IF_ you have `application.properties` and you have a config properties bean then the bean will be filled with the values from the properties
- BUT if you use the config properties bean to modify the values then they stay modified inside the bean only without env properties modifications.
https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config

For me this is a bug since I have to care about two sources of authority in my code: `env` properties and Java bean properties. 

Let's don't mix the different issues together: it's not necessary to generate the Java class and produce an issue with it's usage.
You may generate a Json file instead and forget about all Java related issues in Spring.

If you want to have this kind of functionality then you may use `Environment` bean to set properties into it instead of `System`.
`System` properties should not be touched at all : this is incorrect.

configurationproperties-annotationprocessor/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+
+        <dependency>
+            <groupId>com.github.javaparser</groupId>
+            <artifactId>javaparser-symbol-solver-core</artifactId>
+            <version>3.16.1</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
+
+</project>
```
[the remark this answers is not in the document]
miguelatvaadin: ok. Thanks

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
miguelatvaadin: This code is not from me. It came when I merged the master branch

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
miguelatvaadin: ok. I didn't remember that method

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
miguelatvaadin: ok. thanks

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
miguelatvaadin: Yes. My fault. Changed

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
miguelatvaadin: I think my answer has disappeared, here (and it was a long one). I will try to rewrite it:

As far as I have seen the vaadin servlet reads the initial parameters from the `DeploymentConfiguration` class. It reads the values either from web.xml or from the system properties. Hence why I set them as system properties.

Also, please note that the expected name for each property can be different than the property name set by the user, due to the relaxing naming scheme from springboot. Hence why I do this, to warranty that property names are set as expected.

When I faced this point, I also managed the option of serializing this object to a property map, so it could be passed to `ServletRegistrationBean::setInitparameters` inside `SpringBootAutoConfiguration::servletRegistrationBean`. That would avoid setting system properties, though I discarded this option as I thought it would be less intrusive the other way.

configurationproperties-annotationprocessor/src/main/java/com/vaadin/flow/spring/ConfigurationPropertiesAnnotationProcessor.java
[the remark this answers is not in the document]
miguelatvaadin: Sorry Denis, I have seen your comment now.

I used system properties as I was not sure that, if I set them using `Environment`, they would be actually read by the vaadin servlet.

I chose a java class instead of a json file as I thought that the relaxed naming scheme (which is one of the requirements) was not going to work with a json file.

Regarding the system properties, another option would be to serialize the generated `VaadinConfigurationProperties` object to a properties map, and pass it to `ServletRegistrationBean::setInitparameters` inside  `SpringBootAutoConfiguration::servletRegistrationBean`.

Please note that I need this serialization to warranty that properties are passed using the exact names which the vaadin servlet is expecting.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringBootAutoConfiguration.java
denis-anisimov: The code may not be here if you haven't changed the code: the PR shows a difference between the `master` branch and your branch.

So it's either something wrong with the PR/branch or the branch is out of sync. You should rebase it on top of the master.

denis-anisimov:
Almost all my latest comments are not resolved .
The most important things are:

- dots in the property names:https://github.com/vaadin/spring/pull/635#discussion_r660451632
- dealing with `System` properties  : https://github.com/vaadin/spring/pull/635#discussion_r660465333

Spring app properties should not leak as System properties: this is unexpected.

This PR tries to solve something which is totally different from  original purpose : provide IDE assistance for the Vaadin properties.

The generated class is not used from anywhere: that's fine. It's out of scope of this PR.
This PR should provide IDE assistance: if it introduces an issue then this should be solved in other way.

As I understand `VaadinConfigurationProperties` is created based on `application.properties` : so if properties are defined (in the file ) then the bean will be filled with the values, no need to set `System` properties or whatever: the properties will be read directly from the `Environment` in the existing code and set as initial parameters.
If there are issues with relaxing then it should be fixed there: not via setting the system properties.
So there is no need to read  `VaadinConfigurationProperties` directly : moreover : `VaadinConfigurationProperties` is available _ONLY_ in Spring boot app. The same properties can be used without Spring boot (in Spring MVC)  where `VaadinConfigurationProperties`  doesn't work at all. This is why the fix for https://github.com/vaadin/spring/issues/478 is not eccepted.

PR #642, opened 2020-07-15
Minor class path scanning optimization

### LINKED ISSUE #571: Slow classpath scanning for Spring Boot application when running Application.java
To reproduce:
1. Activate Java 8 in IDE
2. Download [spring starter](https://github.com/vaadin/skeleton-starter-flow-spring)
3. Run `Application.java` 
4. Note that classpath scanning takes a long time, e.g.:
```Search for subclasses and classes with annotations took 5744 ms```
5. Stop the application
6. Run instead the maven goal `spring-boot:run`
7. Note that classpath scanning is faster,e.g.:
```Search for subclasses and classes with annotations took 2656 ms```
8. Activate Java 13 in IDE
9. Note that when running the application in either way the classpath scanning takes the same time:
```Search for subclasses and classes with annotations took 1747 ms```
```Search for subclasses and classes with annotations took 1809 ms```

Expectation: There is no major discrepancy between classpath scanning when running `Application.java` compared to using Maven. As a developer I prefer the former as it enables IDE debugging.

Versions:
- Vaadin 14.2 (or newer)
- IntelliJ IDEA (Ultimate 2019.13)
- Java 8 (jdk1.8.0_202.jdk) 
- Java 13 (openjdk-13.0.2.jdk)

### PR DESCRIPTION
**Pull request by joheriks**
Extend the default class scanning exclusion list to JDK and IDE classes.
Use prefix tree for faster lookups in long exclusion/inclusion lists.
Fixes #571.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.spring;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Quick prefix lookup for package inclusion and exclusion lists.
+ */
+class PrefixTree implements Serializable {
+
+    private final Node root;
+
+    PrefixTree(Collection<String> prefixes) {
+        root = new Node();
+        root.terminal = false;
+        prefixes.forEach(this::addPrefix);
+    }
+
+    void addPrefix(String prefix) {
+        if (prefix.isEmpty()) {
+            throw new IllegalArgumentException("empty prefix");
+        }
+        root.addPrefix(prefix);
+    }
+
+    boolean hasPrefix(String s) {
+        Node node = root;
+        final int slen = s.length();
+        int sidx = 0;
+        while (node != null) {
+            if (node.terminal) {
+                return true;
+            } else if (sidx < slen) {
+                node = node.children.get(s.charAt(sidx++));
+            } else {
+                return false;
+            }
+        }
+        return false;
+    }
+
+    static class Node implements Serializable {
+        private final Map<Character, Node> children = new HashMap<>();
+        private boolean terminal = true;
+
+        void addPrefix(String prefix) {
+            assert !prefix.isEmpty();
+            terminal = false;
+            char ch = prefix.charAt(0);
+            if (!children.containsKey(ch)) {
+                children.put(ch, new Node());
+            }
+            if (prefix.length() > 1) {
+                children.get(ch).addPrefix(prefix.substring(1));
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+
+    static class Node implements Serializable {
+        private final Map<Character, Node> children = new HashMap<>();
+        private boolean terminal = true;
+
+        void addPrefix(String prefix) {
+            assert !prefix.isEmpty();
+            terminal = false;
+            char ch = prefix.charAt(0);
+            if (!children.containsKey(ch)) {
```
mshabarov, line 67: Can `putIfAbsent` be used here?

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        return false;
+    }
+
+    static class Node implements Serializable {
+        private final Map<Character, Node> children = new HashMap<>();
+        private boolean terminal = true;
+
+        void addPrefix(String prefix) {
+            assert !prefix.isEmpty();
```
mshabarov, line 64: Do we need to check the empty prefix if it's already checked within PrefixTree:: addPrefix?

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Quick prefix lookup for package inclusion and exclusion lists.
+ */
+class PrefixTree implements Serializable {
```
mshabarov, line 26: Random thoughts: Have you estimated the profit in performance of using `PrefixTree` as a replacement to `startsWith`? Just wondering. My concern that the `PrefixTree` solution may bring a memory consumption issues. However it's made only at startup.

vaadin-spring/src/test/java/com/vaadin/flow/spring/PrefixTreeTest.java
```diff
@@ -0,0 +1,51 @@
...
+    public void hasPrefix_containsPrefix_returnsTrue() {
+        PrefixTree prefixTree = new PrefixTree(
+                Arrays.asList("com/sun", "antlr", "ch/quos/logback"));
+        Assert.assertTrue(prefixTree.hasPrefix("antlr"));
+        Assert.assertTrue(prefixTree.hasPrefix("com/sun/test"));
+        Assert.assertTrue(prefixTree.hasPrefix("com/sun"));
+    }
+
+    @Test
+    public void hasPrefix_containsPrefix_returnsFalse() {
```
mshabarov, line 36: `hasPrefix_containsPrefix_returnsFalse` -> `hasPrefix_doesNotContainPrefix_returnsFalse`

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
joheriks: Profiling on the default (small) list of exclusions indicated only a minor performance improvement of around 15 %, however the previous solution with `startsWith` performed asymptotically poor because it had time complexity `N * M` where `N` is number of classes and `M` the number of packages in the list. With a prefix tree, this is reduced to on average `N * log(M)`. I do no estimate the memory overhead to be too big. A prefix tree may actually save memory in cases where the inclusion / exclusion lists contain many items with a common prefix.

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
joheriks: Good idea, done.

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
joheriks: Not necessary, removed assertion.

vaadin-spring/src/test/java/com/vaadin/flow/spring/PrefixTreeTest.java
joheriks: Done.

vaadin-spring/src/main/java/com/vaadin/flow/spring/PrefixTree.java
mshabarov: Alright, understood. Looks like a good profit! Thanks

PR #644, opened 2020-07-16
Added Spring Data helper class

### LINKED ISSUE #643: Add a helper method to convert Vaadin specific sort information into Spring Sort
This helper can be made in a generic way, which would help many developers to create sortable lazy data binding to Vaadin Grid.

Spring Data dependencies should be declared optional, so that the dependency is not pulled in by default.

mstahv:
The class contains just one method to make it easier to convert Vaadin sort hints from Query object to Sprind Data Sort, which is used by many backend implementations.

closes #643

### DIFF
--- a/vaadin-spring/pom.xml
+++ b/vaadin-spring/pom.xml
@@ -67,7 +67,17 @@
             <artifactId>spring-boot-autoconfigure-processor</artifactId>
             <optional>true</optional>
         </dependency>
-
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-commons</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-data</artifactId>
+            <optional>true</optional>
+        </dependency>
+        
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
@@ -36,7 +36,7 @@ public interface VaadinSpringDataHelpers {
      * component to a Spring Data based back-end. The method expects Vaadin sort 
      * data to include the property name.
      *
-     * @param vaadinQuery the Vaadin Query object passed by he component
+     * @param vaadinQuery the Vaadin Query object passed by the component
      * @return the Sort object that can be passed for Spring Data based back-end
      */
     static Sort toSpringDataSort(Query<?, ?> vaadinQuery) {
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
@@ -36,16 +36,17 @@ public class VaadinSpringDataHelpers implements Serializable {
      * component to a Spring Data based back-end. The method expects Vaadin sort 
      * data to include the property name.
      *
-     * @param vaadinQuery
-     * @return
+     * @param vaadinQuery the Vaadin Query object passed by he component
+     * @return the Sort object that can be passed for Spring Data based back-end
      */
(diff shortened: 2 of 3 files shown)


vaadin-spring/pom.xml
```diff
@@ -67,7 +67,17 @@
...
-
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-commons</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-data</artifactId>
+            <optional>true</optional>
```
mshabarov, line 78: Minor question: is that dependency have to be an optional indeed? It mean that the developer would have to add it explicitly in his project. Wouldn't it break a concept of using bill of materials (which has all of the Vaadin deps in one place). I don't have a strong experience with `optional`, thus just curious.

vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
```diff
@@ -0,0 +1,51 @@
...
+     * object.
+     * <p>
+     * Can be used as a helper when making a lazy data binding from a Vaadin
+     * component to a Spring Data based back-end. The method expects Vaadin sort 
+     * data to include the property name.
+     *
+     * @param vaadinQuery
+     * @return
+     */
+    public static Sort toSpringDataSort(Query<?, Void> vaadinQuery) {
```
mshabarov, line 42: Why the query filter is `Void`?  I believe this helper must be able to be applied for filtered queries also.

vaadin-spring/pom.xml
mstahv: Flow data is drawn in by all components that use it, so it can be optional here. No developer would use this method without having Vaadin component using flow-data in their project. In theory one could use flow without any components and then flow-data wouldn't be needed. In that case, people would need to exclude flow-data if they want to get rid of obsolete dependencies.

With the Spring Data dependency the optional is more relevant.

Sometimes when optional is used a lot, it is a sign of in-proper modularisation. We could have vaadin-spring-data module, where we'd have only this one method. But I'd not take this route yet for the sake of one helper method.

mshabarov: Understood, thanks!

vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Translates given Query object from a Vaadin component to Spring Data Sort
+     * object.
+     * <p>
+     * Can be used as a helper when making a lazy data binding from a Vaadin
+     * component to a Spring Data based back-end. The method expects Vaadin sort 
+     * data to include the property name.
+     *
+     * @param vaadinQuery the Vaadin Query object passed by he component
```
mshabarov, line 39:
```suggestion
     * @param vaadinQuery the Vaadin Query object passed by the component
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.Serializable;
+import java.util.stream.Collectors;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.domain.Sort.Order;
+
+/**
+ * This class contains helper methods to work with Spring Data based back-ends
+ * and Vaadin components.
+ */
+public class VaadinSpringDataHelpers implements Serializable {
```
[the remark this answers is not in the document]
mstahv: Should maybe be an interface since Java8 🤔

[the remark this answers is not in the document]
mstahv: Changed into interface. Don't really know what is the best practice these days.

vaadin-spring/src/main/java/com/vaadin/flow/spring/data/VaadinSpringDataHelpers.java
mstahv: Changed into ?

PR #650, opened 2020-08-07
Remove outdated pnpmEnabled property and add a bean to support doted prop

### LINKED ISSUE #634: Spring boot configuration vaadin.pnpm-enabled

denis-anisimov:
Fixes #634

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java
@@ -15,19 +15,18 @@
  */
 package com.vaadin.flow.spring;
 
-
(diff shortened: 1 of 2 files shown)


pleku:
This needs to be cherry-picked to 14 and 17, maybe 16 too even though it might not be released for it.

jcgueriaud1:
Was it cherry picked for Vaadin 14 ?
Vaadin spring 12.3.2  (for 14.4.2) looks unchanged.

PR #651, opened 2020-08-07
Read properties using uniform name along with camel case

### LINKED ISSUE #637: Vaadin Properties should be read from the environment using the Spring uniform name
According to this documentation:
`If you read properties from the environment in your application, you will now need to use the uniform name for the property.`

https://github.com/spring-projects/spring-boot/wiki/relaxed-binding-2.0#reading-properties-from-the-environment

This means that vaadin.productionMode is incorrect and should not be read by this add-on.
`env.getProperty("vaadin.productionMode")`

It should be read this way:
`env.getProperty("vaadin.production-mode")`

In practice Spring will bound the property by removing any special characters and converting to lowercase. For example, the following properties all result in the mapping 
vaadin.productionmode=true :
```
vaadin.production-mode=true
vaadin.productionMode=true
vaadin.productionMODE=true
```
The recommendation is that properties are stored in lowercase kabab format. i.e. vaadin.production-mode=true. 
So it won't introduce a breaking change and will reduce the chance or typo error for the user.

All the properties read from Vaadin flow should follow this rules.

This issue is linked to this one: https://github.com/vaadin/spring/issues/634 which asks vaadin property metadata.

denis-anisimov:
Fixes #637

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
@@ -22,6 +22,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.Properties;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringServlet.java
```diff
@@ -154,7 +153,11 @@ private void setProperty(String property, Properties properties) {
     private void setProperty(String envProperty, String initParam,
             Properties properties) {
         Environment env = context.getBean(Environment.class);
-        String value = env.getProperty(envProperty);
+        String value = env
+                .getProperty(SharedUtil.camelCaseToDashSeparated(envProperty));
```
caalador, line 157: Should the dashseparated value be put to lower case? as the string `vaadin.productionMODE` would become `vaadin.production-mODE` but the environment stores all in lower case.

denis-anisimov: Hmm.....
You are right.
I missed that.

But `vaadin.productionMODE` should become `vaadin.production-m-o-d-e`, not `vaadin.production-mode` .
I will correct the code

caalador (approved):

PR #655, opened 2020-08-26
Information and doclink on routes not found error about known Spring Boot Devtools issue 

joheriks:
The issue is https://github.com/spring-projects/spring-boot/issues/19543 and link to workaround.
Related to https://github.com/vaadin/flow/issues/7822

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -86,6 +86,7 @@ import com.vaadin.flow.server.startup.WebComponentConfigurationRegistryInitializ
 import com.vaadin.flow.server.startup.WebComponentExporterAwareValidator;
 import com.vaadin.flow.server.webcomponent.WebComponentConfigurationRegistry;
 import com.vaadin.flow.spring.VaadinScanPackagesRegistrar.VaadinScanPackages;
+import com.vaadin.flow.spring.router.SpringRouteNotFoundError;
 import com.vaadin.flow.theme.Theme;
 
 /**
@@ -631,9 +632,12 @@ public class VaadinServletContextInitializer
 
     private Collection<String> getErrorParameterPackages() {
         return Stream
-                .concat(Stream
-                        .of(HasErrorParameter.class.getPackage().getName()),
-                        getDefaultPackages().stream())
+                .concat(Stream.of(
+                        SpringRouteNotFoundError.class.getPackage().getName()),
+                        Stream.concat(
+                                Stream.of(HasErrorParameter.class.getPackage()
+                                        .getName()),
+                                getDefaultPackages().stream()))
                 .collect(Collectors.toSet());
     }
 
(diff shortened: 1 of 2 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -631,9 +632,12 @@ public void onStartup(ServletContext servletContext)
 
     private Collection<String> getErrorParameterPackages() {
         return Stream
-                .concat(Stream
-                        .of(HasErrorParameter.class.getPackage().getName()),
-                        getDefaultPackages().stream())
+                .concat(Stream.of(
+                        SpringRouteNotFoundError.class.getPackage().getName()),
```
mshabarov, line 636: Why don't you add this new package like this:

```
Stream.of(HasErrorParameter.class.getPackage().getName(), SpringRouteNotFoundError.class.getPackage().getName())
```
?

caalador (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/router/SpringRouteNotFoundError.java
```diff
@@ -0,0 +1,28 @@
+package com.vaadin.flow.spring.router;
+
+import com.vaadin.flow.component.Html;
+import com.vaadin.flow.router.BeforeEnterEvent;
+import com.vaadin.flow.router.ErrorParameter;
+import com.vaadin.flow.router.NotFoundException;
+import com.vaadin.flow.router.RouteNotFoundError;
+
+public class SpringRouteNotFoundError extends RouteNotFoundError {
```
mshabarov, line 9: Do I understand correctly that this message will be shown by default when the route not found error occurs?

caalador: It should override the RouteNotFoundError.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
joheriks: Thanks, that's cleaner (I forgot that `Stream.of` is variadic).

vaadin-spring/src/main/java/com/vaadin/flow/spring/router/SpringRouteNotFoundError.java
joheriks: Yes, that's correct. It will be picked over the other default one from Flow `RouteNotFoundError`.

mshabarov: Ok, understood, thanks!

mshabarov (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/router/SpringRouteNotFoundError.java
mshabarov: Just one more clarification: is that correct that the message from `RouteNotFoundError` would be overridden? What if we would have a route error just because some another problem than that one with Spring bug?

caalador: the int retval = super.setError... adds the original content

mshabarov: Now I see, thanks!

PR #657, opened 2020-08-28
Fix regression with add-on customized RouteNotFoundError subclass

joheriks:
A regression in #655 caused Flow to throw an exception when validating that the route registry had at most one component mapped to `com.vaadin.flow.router.NotFoundException`. Replace `RouteNotFoundError` from Flow with `SpringRouteNotFoundError` instead of adding it when setting up route registry. Custom exception views can inherit from either default.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -293,15 +293,13 @@ public class VaadinServletContextInitializer
                     .getInstance(new VaadinServletContext(
                             event.getServletContext()));
 
-            Stream<Class<? extends Component>> hasErrorComponents = findBySuperType(
+            Set<Class<? extends Component>> errorComponents = findBySuperType(
                     getErrorParameterPackages(), HasErrorParameter.class)
                             .filter(Component.class::isAssignableFrom)
                             // Replace Flow default with custom version for Spring
                             .filter(clazz -> clazz != RouteNotFoundError.class)
-                            .map(clazz -> (Class<? extends Component>) clazz);
-            final Set<Class<? extends Component>> errorComponents =
-                    hasErrorComponents.collect(Collectors.toSet());
-
+                            .map(clazz -> (Class<? extends Component>) clazz)
+                    .collect(Collectors.toSet());
             if (errorComponents.stream().noneMatch(RouteNotFoundError.class::isAssignableFrom)) {
                 errorComponents.add(SpringRouteNotFoundError.class);
             }
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -42,6 +42,8 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import com.googlecode.gentyref.GenericTypeReflector;
+
+import com.vaadin.flow.router.RouteNotFoundError;
 import com.vaadin.flow.server.startup.ServletDeployer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -294,9 +296,16 @@ public void failFastContextInitialized(ServletContextEvent event) {
...
+                            // Replace Flow default with custom version for Spring
+                            .filter(clazz -> clazz != RouteNotFoundError.class)
                             .map(clazz -> (Class<? extends Component>) clazz);
-            registry.setErrorNavigationTargets(
-                    hasErrorComponents.collect(Collectors.toSet()));
+            final Set<Class<? extends Component>> errorComponents =
+                    hasErrorComponents.collect(Collectors.toSet());
+
+            if (errorComponents.stream().noneMatch(RouteNotFoundError.class::isAssignableFrom)) {
+                errorComponents.add(SpringRouteNotFoundError.class);
```
mshabarov, line 306: Do I understand correctly, that if the application has a custom `RouteNotFoundError` view (subclass), then this spring view is never shown?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -294,9 +296,16 @@ public void failFastContextInitialized(ServletContextEvent event) {
             Stream<Class<? extends Component>> hasErrorComponents = findBySuperType(
                     getErrorParameterPackages(), HasErrorParameter.class)
                             .filter(Component.class::isAssignableFrom)
+                            // Replace Flow default with custom version for Spring
+                            .filter(clazz -> clazz != RouteNotFoundError.class)
```
mshabarov, line 300: We could probably combine those two `filter` calls into one call. I can't recall if those two filter calls lead to items traverse running twice, but at least it won't create extra Stream object.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -294,9 +296,16 @@ public void failFastContextInitialized(ServletContextEvent event) {
...
                             .map(clazz -> (Class<? extends Component>) clazz);
-            registry.setErrorNavigationTargets(
-                    hasErrorComponents.collect(Collectors.toSet()));
+            final Set<Class<? extends Component>> errorComponents =
+                    hasErrorComponents.collect(Collectors.toSet());
+
+            if (errorComponents.stream().noneMatch(RouteNotFoundError.class::isAssignableFrom)) {
+                errorComponents.add(SpringRouteNotFoundError.class);
+            }
+            registry.setErrorNavigationTargets(errorComponents);
```
mshabarov, line 308: I would also create a test for this regression, to be safe.

mshabarov (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
pleku: I read it like that too, and that is probably how it should be. 
The `SpringRouteNotFoundError` is useful for those who suffer from the Spring dev tools issue initially. I'd suspect that when people add their own error views, they are past the point when they need the instructions. Unless they start by taking a starter like Bakery which overrides the view.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
pleku: I agree, either unit test or manual testing needed. Can this be unit tested ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
joheriks: Yes, the default message should disappear after being overidden. Bakery could add the same message to its customized view as well if deemed necessary (though probably it isn't).

pleku: BTW related to the original change, it doesn't anywhere seem to check whether or not the application is in development/production mode. Is it so that the default error view is not even shown in production mode or, is it now outputting the error message even though it is in production mode and where it doesn't make sense for the end user ?

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
joheriks: Unit test added.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
joheriks: Yeah, prod mode should of course be checked 🤦‍♂️. Fixed.

pleku (approved):

PR #664, opened 2020-10-01
fix: custom pushURL failed with non-root mapping

joheriks:
When mapping was `/*`, `SpringBootAutoConfiguration` updated init parameter `pushURL` for `SpringServlet` but this was overridden later if custom `pushURL` was provided by user. Fixed by moving the push path update logic to just before deployment config is created.
Fixes vaadin/flow#9077.

### DIFF
--- a/vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
+++ b/vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
@@ -33,8 +33,7 @@ import com.vaadin.flow.shared.communication.PushMode;
 import com.vaadin.flow.spring.instantiator.SpringInstantiatorTest;
 
 @RunWith(SpringRunner.class)
-@SpringBootTest(properties = { "vaadin.push-mode=MANUAL",
-        "vaadin.push-u-r-l=someUrl" })
+@SpringBootTest(properties = { "vaadin.push-mode=MANUAL" })
 @Import(TestServletConfiguration.class)
 public class SpringServletTest {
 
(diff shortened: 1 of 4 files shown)


pleku:
Build failed due to compile error ?

vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
```diff
@@ -52,4 +52,13 @@ public void readUniformNameProperty_propertyNameContansDash_propertyNameIsConver
         Assert.assertEquals("someUrl",
                 service.getDeploymentConfiguration().getPushURL());
     }
+
+    @Test
+    public void customPushUrl_rootMapping_isPrefixedWithContextVaadinMapping()
```
pleku, line 57: I know that this didn't even have any test before, but could maybe have another unit test just for the introduced mapping logic with push urls like "/someUrl" and apparently "/vaadinServlet/someUrl".

An IT would be great but it would be rather complex to make happen so I think it can be skipped.

pleku (changes requested):

vaadin-spring/src/test/java/com/vaadin/flow/spring/service/SpringServletTest.java
joheriks: Added unit tests for no custom push URL, custom URL with and without preceding slash, as well as custom URL with preceding `/vaadinservlet`.

joheriks:
> Build failed due to compile error ?

Due to the previously [missing polymertemplate import](https://github.com/vaadin/spring/pull/663), now merged so validation should pass.

pleku (approved):

PR #665, opened 2020-10-02
chore: bump version to 15.0 for v18

pleku:
Needed because Flow version has bumped a major and there are significant
changes. Flow version used in build is already 5.0-SNAPSHOT.


joheriks (approved):
Perhaps README.md could be updated to state that 14.0 is now for Vaadin 17 and 15.0 is for upcoming majors.

PR #708, opened 2020-12-23
fix: get rid of deployment config in servlet context initializer

denis-anisimov:
* Implement a Spring application config factory
* Use the Spring config app factory in Spring boot case
* Adapt unit tests to use the proper config


caalador (commented):
Seems fine to me. Soroosh can have a look also as I had not that much time.

vaadin-spring/src/test/java/com/vaadin/flow/spring/SpringStubServletConfigTest.java
```diff
@@ -1,84 +0,0 @@
...
-        when(registration.getServletName()).thenReturn("vaadinServlet");
-        when(context.getInitParameterNames())
-                .thenReturn(Collections.emptyEnumeration());
-        when(registration.getInitParameters())
-                .thenReturn(Collections.emptyMap());
-
-    }
-
-    @Test
-    public void vaadinProductionMode_isReadFromEnvironmentVariables() {
```
pleku, line 64: Non-blocking question: Would this have made any sense anymore as another test without the stub config ?

pleku (dismissed):
Just one question, otherwise LGTM

vaadin-spring/src/test/java/com/vaadin/flow/spring/SpringStubServletConfigTest.java
pleku: If not, good riddance. If yes, then can be added back later (to get this in)

denis-anisimov: There is already a test which replaces this one.
Except this one checks only one single property `productionMode`.
`SpringApplicationConfigurationFactoryTest`  tests all `SpringServlet.PROPERTY_NAMES` properties.
And `SpringApplicationConfigurationFactory` is the proper SPI impl replacement for ugly stub config

pleku (approved):

Issue #707, opened 2020-12-23
Make Spring ApplicationConfigurationFactory which can be overriden


denis-anisimov:
Spring add-on requires its own ApplicationConfigurationFactory because the `ApplicationConfiguration` should be based on Spring application properties.

The work for Spring `ApplicationConfigurationFactory` is in progress and will be done soon.
**But** there will be no way for the Spring app developer to override  provided Spring `ApplicationConfigurationFactory` because Flow Lookup SPI requires only one service impl in the classpath.
Also the implemented Spring `ApplicationConfigurationFactory` won't be a Spring bean (the service is instantiated directly via default CTOR).

All these issues should be fixed.
But to be able to make it properly the fix for https://github.com/vaadin/flow/issues/9717 in Flow should be done.

denis-anisimov:
related #710

denis-anisimov:
Fixed by https://github.com/vaadin/spring/pull/711

PR #711, opened 2021-01-06
Use lookup initializer SPI to create and init Lookup using Spring aware beans

### LINKED ISSUE #710: SpringApplicationConfigurationFactory sometimes may not use ApplicationContext
`ApplicationContext` may be not yet available when `SpringApplicationConfigurationFactory` is executed.

That happens when the app is run as WAR.
Apparently there is a race condition between plain `ServletContextinitializer`  used in Flow and web initializer in Spring.
If Spring web initializer is run first then it sets the app context.
If it's run after then context is not yet availble.

Their execution order is non deterministic and there is no way to get app context in `SpringApplicationConfigurationFactory` reliably.

 `SpringApplicationConfigurationFactory` is used to set config properties based on Spring configuration.
This code is a replacement for Stub servlet config which has been invalidly used in `VaadinServletContextInitializer`.
Apparently the need of this code is to have Spring properties in the Flow `ServletContextinitializer`s   (like dev mode initializer).

This code works when Spring is executed as a Java Spring boot application because in that case `VaadinServletContextInitializer` runs all its internal `ServletContextListener`s.
And this code apparently has never been working in case of WAR Vaadin app: the application context me be not available at the time when e.g. `DevModeInitializer` has been executed and this initializer is defined in Flow (not in Spring add-on) and it's executed once so the code inside `VaadinServletContextInitializer` doesn't do anything.

So this is not a new issue.
This has been broken for ages in case WAR.
(`VaadinServletContextInitializer` makes an illusion that it always executed everything . In fact it's needed _ONLY_ in Spring boot Java main app and this is a huge issue).

But this should be solved at some point. For now I've just added a check whether the app context is available and super impl is returned if it's not available. 

This ticket is related to https://github.com/vaadin/spring/issues/707
and https://github.com/vaadin/flow/issues/9717.

It should be fixed once these tickets are done.

denis-anisimov:
fixes #710, #707


vaadin-spring-tests/test-ts-services/src/test/java/com/vaadin/flow/connect/AppViewIT.java
```diff
@@ -29,6 +30,7 @@
 /**
  * Class for testing issues in a spring-boot container.
  */
+@Ignore
```
caalador, line 33: Could you add a comment for the ignore naming ticket 9751

vaadin-spring-tests/test-ts-services-custom-client/src/test/java/com/vaadin/flow/connect/AppViewIT.java
```diff
@@ -30,6 +30,7 @@
 /**
  * Class for testing issues in a spring-boot container.
  */
+@Ignore
```
caalador, line 33: Could you add a comment for the ignore naming ticket 9751

caalador (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicaitonConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>
+ * Registers a default {@link ApplicationConfigurationFactory} for Vaadin web
+ * application if there is no developer provided factory available.
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+@Configuration
+public class VaadinApplicaitonConfiguration {
```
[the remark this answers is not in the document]
denis-anisimov: Done.

vaadin-spring-tests/test-ts-services/src/test/java/com/vaadin/flow/connect/AppViewIT.java
denis-anisimov: Done

vaadin-spring-tests/test-ts-services-custom-client/src/test/java/com/vaadin/flow/connect/AppViewIT.java
denis-anisimov: Done

caalador (dismissed):

caalador (approved):

PR #722, opened 2021-02-23
feat: add a helper for configuring Spring Security for Vaadin

### LINKED ISSUE #678: add a helper to configure Spring Security for Vaadin
When using Vaadin in a Spring Boot application with Spring Security, I want to have helper methods like `VaadinSpringSecurity.configure(http)` and `VaadinSpringSecurity.configure(web)` so that configuring spring security is a one-liner as opposed to the current best practice recommendation of:
```java
http.authorizeRequests()
        // allow anonymous access to the index page and static assets
        .antMatchers("/", 
        "/VAADIN/**", 
        "/vaadinServlet/**", 
        "/images/**", 
        "/icons/**", 
        "/manifest.webmanifest",
        "/sw.js",
        "/offline.html").permitAll()
        // all other requests
        .anyRequest().authenticated()
        .and().oauth2ResourceServer().jwt();
```

joheriks:
Fixes #678 

Depends on https://github.com/vaadin/flow/pull/10134


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinSpringSecurity.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Matcher for Vaadin static resources.
+     *
+     * @return default {@link WebSecurity} ignore matcher
+     */
+    public static RequestMatcher getDefaultWebSecurityIgnoreMatcher() {
+        return new OrRequestMatcher(Stream.of(
+                "/VAADIN/**",
```
knoobie, line 68: /VAADIN/ ist dangerous, see https://github.com/vaadin/spring/issues/678#issuecomment-736337987 and the comments above.

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinSpringSecurity.java
```diff
@@ -0,0 +1,105 @@
...
+    public static RequestMatcher getDefaultWebSecurityIgnoreMatcher() {
+        return new OrRequestMatcher(Stream.of(
+                "/favicon.ico",
+                "/images/**",
+                "/icons/**",
+                "/manifest.webmanifest",
+                "/sw.js",
+                "/sw-runtime-resources-precache.js",
+                "/offline.html")
+                .map(AntPathRequestMatcher::new)
```
knoobie, line 76: I'm not sure if you call this static strings now "public API" if yes, for every case should be a test that those aren't removed.

joheriks: Done, added a case to exercise each matcher. Also added test that non-matching resource is not accessible.

platosha (approved):

PR #744, opened 2021-04-01
Provide a default request cache which ignores non-route requests

Artur-:
Fixes https://github.com/vaadin/flow/issues/10344
Depends on https://github.com/vaadin/flow/pull/10363

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
@@ -1,0 +1,41 @@
+package com.vaadin.flow.spring.security;
+
+import javax.servlet.http.HttpServletRequest;
+
+import com.vaadin.flow.server.HandlerHelper;
+import com.vaadin.flow.spring.VaadinConfigurationProperties;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Contains utility methods related to request handling.
+ */
+@Component
+public class RequestUtil {
+
+    @Autowired
+    private VaadinConfigurationProperties configurationProperties;
+
+    /**
+     * Checks whether the request is an internal request.
+     *
+     * An internal request is one that is needed for all Vaadin applications to
+     * function, e.g. UIDL or init requests.
+     *
+     * Note that bootstrap requests for any route or static resource requests are
+     * not internal, neither are resource requests for the JS bundle.
+     *
+     * @param request the servlet request
+     * @return {@code true} if the request is Vaadin internal, {@code false}
+     *         otherwise
+     */
+    public boolean isFrameworkInternalRequest(HttpServletRequest request) {
+        String vaadinMapping = configurationProperties.getUrlMapping();
+        if (vaadinMapping.endsWith("/*")) {
+            vaadinMapping = vaadinMapping.substring(0, vaadinMapping.length() - "/*".length());
+        }
+        return HandlerHelper.isFrameworkInternalRequest(vaadinMapping, request);
+    }
+
+}
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
@@ -1,0 +1,55 @@
+package com.vaadin.flow.spring.security;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.vaadin.flow.server.connect.EndpointUtil;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
+import org.springframework.stereotype.Component;
+
+/**
+ * A default request cache implementation which aims to ignore requests that are
+ * not for routes.
+ * 
+ * Using this class helps with redirecting the user to the correct route after
+ * login instead of redirecting to some internal URL like a service worker or
+ * some data the service worker has fetched.
+ */
+@Component
+public class VaadinDefaultRequestCache extends HttpSessionRequestCache {
+
+    @Autowired
+    private EndpointUtil endpointUtil;
+
+    @Autowired
+    private RequestUtil requestUtil;
+
+    @Override
+    public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
+        if (requestUtil.isFrameworkInternalRequest(request)) {
+            return;
+        }
+        if (endpointUtil.isEndpointRequest(request)) {
+            return;
+        }
+        if (isServiceWorkerInitiated(request)) {
+            return;
+        }
+
+        super.saveRequest(request, response);
+    }
+
+    /**
+     * Checks if the request is initiated by a service worker.
+     * 
+     * NOTE This method can never be used for security purposes as the "Referer"
+     * header is easy to fake.
+     */
+    private boolean isServiceWorkerInitiated(HttpServletRequest request) {
+        String referer = request.getHeader("Referer");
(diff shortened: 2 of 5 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -0,0 +1,41 @@
...
+     * function, e.g. UIDL or init requests.
+     *
+     * Note that bootstrap requests for any route or static resource requests are
+     * not internal, neither are resource requests for the JS bundle.
+     *
+     * @param request the servlet request
+     * @return {@code true} if the request is Vaadin internal, {@code false}
+     *         otherwise
+     */
+    public boolean isFrameworkInternalRequest(HttpServletRequest request) {
```
haijian-vaadin, line 33: Could have a unit test for this public method?

haijian-vaadin (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
haijian-vaadin: Or do we actually need this RequestUtil? the JavaDoc in HandlerHelper says
```
@param servletMappingPath
     *            the path the Vaadin servlet is mapped to, with or without and
     *            ending "/*"
```
If the JavaDoc in https://github.com/vaadin/flow/pull/10363 is true, then we can use HandlerHelper directly without of RequestUtil.

Artur-: You need to pass the servlet mapping that is used in Spring to ensure that a request to another servlet is not interpreted as a Vaadin internal request. I will add a unit test

haijian-vaadin: OK, but if statement is not necessary?
```
if (vaadinMapping.endsWith("/*")) {
    vaadinMapping = vaadinMapping.substring(0, vaadinMapping.length() - "/*".length());
}
```
based on what I read from the code, then "/*" ending is handled already in HandlerHelper?

Artur-: Removed

Artur-: and added unit test

haijian-vaadin (approved):

haijian-vaadin (changes requested):
Can we mark the fusion dependency as provided/optional? For people who don't use Fusion features at all, they may want to exclude the fusion dependencies.

Artur-:
Fusion is now optional as proven by the test project which does not include fusion

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
```diff
@@ -0,0 +1,69 @@
...
+@Component
+public class VaadinDefaultRequestCache extends HttpSessionRequestCache {
+
+    @Autowired
+    private ApplicationContext applicationContext;
+
+    @Autowired
+    private RequestUtil requestUtil;
+
+    private Object endpointUtil;
```
haijian-vaadin, line 31: Don't need to use Object, it can actually be `EndpointUtil endpointUtil`

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @PostConstruct
+    public void init() {
+        try {
+            endpointUtil = applicationContext.getBean("com.vaadin.flow.server.connect.EndpointUtil");
+        } catch (Exception e) {
+            // Presumable Fusion is not on the classpath
+        }
```
haijian-vaadin, line 39: Can we wrap this into a separate method? like
private EndpointUtil getEntpointUtil() {
   try {
            endpointUtil = applicationContext.getBean("com.vaadin.flow.server.connect.EndpointUtil");
            return (EndpointUtil) endpointUtil;
        } catch (Exception e) {
            return null;
            // Presumable Fusion is not on the classpath
        }
}

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
```diff
@@ -0,0 +1,69 @@
...
+
+    }
+
+    @Override
+    public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
+        if (requestUtil.isFrameworkInternalRequest(request)) {
+            return;
+        }
+        if (endpointUtil != null && ((EndpointUtil) endpointUtil).isEndpointRequest(request)) {
```
haijian-vaadin, line 48: Then we don't need the cast here.

haijian-vaadin (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
Artur-: How? It results in `Caused by: java.lang.IllegalStateException: Failed to introspect Class [com.vaadin.flow.spring.security.VaadinDefaultRequestCache] from ClassLoader [sun.misc.Launcher$AppClassLoader@18b4aac2]`

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
Artur-: How? It results in `Caused by: java.lang.IllegalStateException: Failed to introspect Class [com.vaadin.flow.spring.security.VaadinDefaultRequestCache] from ClassLoader [sun.misc.Launcher$AppClassLoader@18b4aac2]`

haijian-vaadin (approved):

haijian-vaadin (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/RequestUtil.java
```diff
@@ -0,0 +1,64 @@
...
+    public void init() {
+        try {
+            endpointUtil = applicationContext.getBean(EndpointUtil.class);
+        } catch (NoClassDefFoundError e) {
+            // Presumable Fusion is not on the classpath
+        } catch (Exception e) {
+            // Presumable Fusion is not on the classpath
+        }
+
+    }
```
joheriks, line 38: Would a cleaner solution be injecting 
```
@Autowired(required=false)
private EndpointUtil endpointUtil;
```

Artur-: That was the initial solution. Then it was changed to this to support running without Fusion on the classpath

Artur-: `required=false` still requires the class to be present

haijian-vaadin: Copy paste a previous reply from Artur 

> It results in Caused by: java.lang.IllegalStateException: Failed to introspect Class [com.vaadin.flow.spring.security.VaadinDefaultRequestCache] from ClassLoader [sun.misc.Launcher$AppClassLoader@18b4aac2]

joheriks: Right. My concern was if there is risk that exceptions in bean creation not being distinguishable from bean missing. But I guess the exception chain is printed anyway in either case.

joheriks (approved):
Tested on bakery in stead of the custom request cache. LGTM.

PR #748, opened 2021-04-07
feat: provide a Spring Security base class

Artur-:
Aims to by default secure the whole application by default and allow overriding configuration to make some views or resources public


haijian-vaadin (changes requested):
Should we also change the [SecurityConfig.java](https://github.com/vaadin/spring/blob/master/vaadin-spring-tests/test-spring-security-helper/src/main/java/com/vaadin/flow/spring/test/SecurityConfig.java) to extend from the VaadinWebSecurityConfigurerAdapter instead?

Artur-:
Fixed

haijian-vaadin (approved):

PR #750, opened 2021-04-07
chore: use the same formatter as in Flow

Artur-:


pom.xml
```diff
@@ -152,6 +151,23 @@
                     </filesets>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>net.revelc.code.formatter</groupId>
+                <artifactId>formatter-maven-plugin</artifactId>
+                <version>2.11.0</version>
+                <configuration>
+                    <configFile>eclipse/VaadinJavaConventions.xml</configFile>
```
joheriks, line 159: Isn't the formatter spec required to be in the repo, or how is it located?

joheriks (changes requested):

pom.xml
Artur-: This is apparently what the flow buildtools dependency provides, see https://github.com/vaadin/flow/blob/master/build-tools/pom.xml#L19

joheriks: Right, `flow-buildtools` includes `../eclipse` in Flow. My bad.

joheriks (approved):

PR #754, opened 2021-04-07
Ignore CSRF for framework and endpoint requests

Artur-:


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+public abstract class VaadinWebSecurityConfigurerAdapter extends WebSecurityConfigurerAdapter {
+
+    @Autowired
+    private VaadinDefaultRequestCache vaadinDefaultRequestCache;
+
+    @Autowired
+    private EndpointUtil endpointUtil;
+
+    @Autowired
+    private RequestUtil requestUtil;
```
knoobie, line 50: I don't think this direct dependency to fusion beans is okay in such a generic class. 

For example: We exclude fusion-endpoint from every application to reduce the huge amount of dependencies it depends on and we don't need as flow only user. Introducing the direct dependency here forces me to have fusion on the class path and even have a bean from fusion. 

Is this really necessary? 

Looking at the investigation in flow about separating flow and fusion, it doesn't really make sense to force spring user to have both of them again. 

CC: @pleku

Artur-: Wouldn't it be enough to mark this as `required=false`?

Sounds like the real problem is in the dependencies of Fusion though. There should be no need to manually optimize like that

knoobie: Using required false with a null check later on could work. 

Cleaning up the dependencies in fusion would be good too but I don't know if that's possible without a deep refactoring or improved production build. If I'm not mistaking we saved like 10-20mb just by excluding fusion from our final jar but I have to check again on my PC later.

knoobie: I got time to check it again:

- V19.0.3 (starter) with `fusion-endpoint` - `mvn clean package -Pproduction` took 43sec and the produced jar is ~54mb
- V19.0.3 (starter) **without** `fusion-endpoint` - `mvn clean package -Pproduction` took 21sec and the produced jar is ~39mb

Exclusion was done like this:

```xml
       <dependency>
            <groupId>com.vaadin</groupId>
            <!-- Replace artifactId with vaadin-core to use only free components -->
            <artifactId>vaadin</artifactId>
            <exclusions>
               <exclusion>
                 <groupId>com.vaadin</groupId>
                 <artifactId>fusion-endpoint</artifactId>
               </exclusion>
            </exclusions>
        </dependency>


```

Artur-: This is fixed in the PR where the code is actually introduced: #744

knoobie: @Artur- You are the best! Thank you :)

Artur-:
https://github.com/vaadin/docs/pull/345 can be merged after this

haijian-vaadin (changes requested):
This is really nice. Could we have a test to verify the change works? e.g. in the `test-spring-security-helper` module?

Artur-:
It should be covered already by
```
AppViewIT.deep_page_should_require_login
AppViewIT.root_page_should_require_login
```
which both fail if you just make the change to the test `SecurityConfig.java`

haijian-vaadin:
> It should be covered already by
> 
> ```
> AppViewIT.deep_page_should_require_login
> AppViewIT.root_page_should_require_login
> ```
> 
> which both fail if you just make the change to the test `SecurityConfig.java`

OK, but I think it only covers CSRF for framework requests, not the endpoint requests. Would be good if we can turn the test app in `test-spring-security-helper` to a hybrid app and also have an endpoint there.

Artur-:
I think we should split `test-spring-security-helper` into `test-spring-security-flow` and `test-spring-security-fusion` but not as part of this PR. Would that work?

knoobie:
^I like that idea - additionally with a `test-spring-security-combined` all use cases can be covered.

haijian-vaadin:
OK, that could work, let's merge the PR.

haijian-vaadin (approved):

PR #758, opened 2021-04-09
Rename test-spring-security-helper -> test-spring-security-flow

Artur-:


knoobie:
Should `fusion` be excluded from the project to showcase that this is flow only and works without `fusion` as dependency?

Artur-:
It is actually excluded right now as `fusion-endpoint` is set as provided in the `vaadin-spring` pom.xml.

knoobie:
You are right, `vaadin-spring` doesn't use `vaadin-core` and defines all dependencies by itself.

haijian-vaadin (dismissed):

vaadin-spring-tests/test-spring-security-flow/pom.xml
```diff
@@ -7,8 +7,8 @@
         <artifactId>vaadin-spring-tests</artifactId>
         <version>17.0-SNAPSHOT</version>
     </parent>
-    <artifactId>test-spring-secturity-helper</artifactId>
-    <name>Integration tests for Vaadin Spring Security helper</name>
+    <artifactId>test-spring-secturity-flow</artifactId>
```
joheriks, line 10: Noticed a typo in the original artifactId (`secturity`) that perhaps could be fixed in this change as well.

Artur-: Ifxed teh ypto

joheriks (approved):

PR #766, opened 2021-04-14
Make Flow security example more real

Artur-:
Allows public access to /
Requires login for /private
Uses a database for login information and account balance
Login as john/john (user) or emma/emma (admin)

Icon must be in src/main/resources/META-INF/resources/icons/icon.png for Flow to pick it up, see https://github.com/vaadin/flow/issues/10632

NOTE: This is still missing access control for the views so you can access /private when not logged in. It will result in an NPE when it should redirect to login instead


Artur-:
Requires https://github.com/vaadin/flow/pull/10700

joheriks (approved):

PR #781, opened 2021-04-26
feat: add setLoginView for Fusion login views

Artur-:
This automatically sets up form login and uses the needed VaadinSavedRequestAwareAuthenticationSuccessHandler

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
@@ -10,6 +10,7 @@ import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.builders.WebSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.config.annotation.web.configurers.ExpressionUrlAuthorizationConfigurer;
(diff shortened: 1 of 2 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
```diff
@@ -108,4 +109,48 @@ public static RequestMatcher getDefaultWebSecurityIgnoreMatcher() {
...
+     * @param http
+     *            the http security from {@link #configure(HttpSecurity)}
+     * @param fusionLoginViewPath
+     *            the path to the login view
+     * @param logoutUrl
+     *            the URL to redirect the user to after logging out
+     * @throws Exception
+     *             if something goes wrong
+     */
+    protected void setLoginView(HttpSecurity http, String fusionLoginViewPath,
```
haijian-vaadin, line 147: is it specific to Fusion or apply to Flow as well? Better to rename it if it applies to both. Also, don't forget the javadoc when renaming.

haijian-vaadin (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
Artur-: It is intended for Fusion. The Flow version in #780 takes a `Component` class instead of a String

haijian-vaadin (approved):

PR #796, opened 2021-05-11
feat: implement RouteScope

### LINKED ISSUE #263: RouteScope support for Vaadin Spring Flow version
See also https://github.com/vaadin/spring/issues/473

What to do : 
https://github.com/vaadin/spring/issues/263#issuecomment-527863137

@stefanuebe commented on [Fri Mar 23 2018](https://github.com/vaadin/flow/issues/3755)

We have an application with several views, that contains grids with huge amounts of data and a lot of components. Parallel to this data there are editor dialogues and sidebars, that show detail data or allows the editing of this data. Dialogues and views are communication via Springs ApplicationEventsPublisher.

The problem here is, that if we use "prototype" as a scope, the event communication does not work since he creates new bean instances for this communication.

On the other hand the "ui" scope leads to an unnecessary usage of memory when switching to other views, since the unused view will not be needed for a time (and since we do not know, if it will be needed again, we can not just remove it). Also they will be still react on events even if not necessary.
 
ViewScope of V8 handled this scenario perfectly (at least in theory). Since this scenario does not seem to be a rare scenario for a Spring based Vaadin application, please bring back the ViewScope.

---

@Legioth commented on [Fri Mar 23 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-375661364)

The reason this hasn't been implemented so far is that nobody has come up with a good definition of exactly when a "view" scope would change. This was (relatively) straightforward with the quite simple one-level `Navigator` architecture in e.g. Vaadin 8, but gets quite messy with the more versatile routing architecture in Flow.

Some interesting examples include
* Should the view scope change when navigating from `/foo/1` to  `/foo/2` when both are handled by `@Route("foo")` that implements `HasUrlParameter`?
* Should the view scope change when navigating from `/foo` to `/bar` if both routes are handled by a class with `@Route("foo") @RouteAlias("bar")`?
* Should the view scope change when navigating from `/foo` to `/bar` if the class with route  `@Route("foo")` also implements `RouterLayout` and is defined as the layout for the `/bar` route?

---

@kumm commented on [Tue Apr 03 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378198099)

I am experiencing with view context on the CDI side, and have some plans what to implement.

Initially bound view context to view target lifecycle controlled by Vaadin navigation. 
When Instantiator::createRouteTarget is called with the route target of the navigation, the scope should change before creation. - Now i am learning what is the answer to your questions under this circumstances. Guess all of them is no. Seems createRouteTarget is called only if target component does not exist in the previous route component chain  -

Next step to introduce view context strategies like in official vaadin CDI addon 3.0. This way it would be configurable per view, when to change scope. - Now it seems to be hard to implement, so next means maybe some time -

Of course there are many details currently i can't see, these are my inital toughts.

---

@kumm commented on [Wed Apr 04 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378398182)

Updated my previous comment. 
Still think let the framework decide. Calling Instantiator with the route target means a scope change. 
Strategies are delayed.

---

@Legioth commented on [Wed Apr 04 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378551946)

One challenge that came to mind is the situation during navigation when old components have not yet been detached, but new components have already been created. This has also been a problem with implementations for Vaadin 8.

The only "pure" solution would be to change Flow to fully get rid of the old components before even starting to create any new ones. Supporting this would slightly complicate some internal logic that is already quite heavy, but it might still be worthwhile.

---

@Legioth commented on [Wed Apr 04 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378553574)

One case that I'm still pondering is when the same component class has `@Route` and is also a `RouterLayout` for other `@Route` classes.

If we were to simplify scope implementations by splitting up `Instantiator.createRouteTarget` into separate methods for route layouts and the actual navigation target, then I'm wondering whether we would also have to refrain from reusing an instances whose role changes? Otherwise, there might be cases when the active navigation target changes, but Flow isn't invoking the method for retrieving a (new) navigation target instance since it reuses an instance that was previously used as a router layout.

---

@stefanuebe commented on [Wed Apr 04 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378569733)

Is there a meaningful use case for having a class having `@Route` and being the router layout at the same time? In v8 it was not possible to be a view and the view display at the same time afaik, maybe this should simply be the same in v10? 

---

@stefanuebe commented on [Wed Apr 04 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378569827)

Sorry, wrong button

---

@kumm commented on [Thu Apr 05 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378752719)

As @stefanuebe currently I am thinking about a good reason to support a layout-and-target in one class.

@Legioth: Indeed removing old components from the tree before destroying their contexts would be great. Now seem to work with pseudo scopes (without client proxy) as in V8, but it should be avoided.

---

@pleku commented on [Thu Apr 05 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378825260)

Layout-and-target is a very common thing to do, for instance when you have a master-details type of view structure, and there isn't always a details opened. So for instance you have a `Grid` in as the master data browser, and on selection you open the form on the side of the grid to edit the data. There I think everyone will use the view with the grid as both a target and a layout.

---

@stefanuebe commented on [Thu Apr 05 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378831038)

How about adding parameters to the annotation to solve some of the problems, @Legioth mentioned in the beginning? 

Lets say we have a `@ViewScope` (or maybe better called `@RouteScope` since it is controlled by `@Route` and not `View` anymore) with optional `boolean` arguments, that control the behavior for such special cases:
- `keepInstanceOnParameterChange` - don't create a new instance when the current route is re-opened with different parameters (so the crucial part of the url doesn't change in the navigation)
- `keepInstanceForAlias` - keep the current instance when the url changes to an alias
- `keepInstanceOnRouteAndLayoutSwitch` - when current instance is a router layout and also a route without a specific router layout and and the user navigates from this instance as a route to this instance as a router layout, there will be no new instance created but the current instance kept.
- ... (other use cases)

So each developer can handle those cases related to their own use case or work flow and we do not have to handle all special ones in a default implementation.
The names of the parameters are of course just suggestions ;)

---

@kumm commented on [Thu Apr 05 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378844887)

@pleku in this case why not create a ```@Route(value="", layout=MyMasterGrigLayout.class)``` with a label 'please select an item', and leave MyMasterGrigLayout to be just a ParentLayout?

Of course it is not a real solution in a DI environment. The master Layout have to be dependent, or UI scoped. 
We could handle the top level route target as the context driver Route target. In this case all child route of the master layout-and-route, and the master itself (the whole subtree) would belong to the same view scope. But since there can be endless level, i don't think one scope is enough. 
It sounds me something like conversation ( or ```@ConversationGroup``` in Deltaspike ) context. 

@stefanuebe it is similar to view context strategies. I've delayed it because i've found it really hard to implement with the current concepts of the navigation in Flow. Currently Flow makes the decision to query, or not to query a route component. Would need to change it ( or would require to reimplement navigation inside an addon ).

---

@Legioth commented on [Thu Apr 05 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-378857409)

I suspect something like `keepInstanceOnRouteAndLayoutSwitch` would be really weird in practice. It would require that the child views could appear into the scope when navigating into the details view and then disappear again when navigating back. Alternatively all potential children would have to be included in the scope all the time. And then things would get even more weird if the same pattern is repeated for multiple levels (i.e. if the child is also a layout-and-target).

I agree that it might make more sense to somehow allow defining any component as a context root, destroying the context when that component is detached. I would assume that the target component would then be identified with a class literal in some annotation, with the requirement that there's only one component of that type attached in `UI.getCurrent()`.

If I correctly understood the `@ConversationGroup` documentation from DeltaSpike, this is technically not a scope, but rather a qualifier that is used for similar purposes as a scope. Would this mean that this could work for injecting beans, but not necessarily for other features that are based on scopes, e.g. event propagation?

---

@kumm commented on [Fri Apr 06 2018](https://github.com/vaadin/flow/issues/3755#issuecomment-379076138)

It was a long time ago I used it with JSF, but I don't remember any missing CDI features. The scope there is ```@GroupedConversationScoped```, and beans are grouped by the qualifier. Event propagation should work regardless observer beans are qualified or not. 

It is not a problem to implement a ```@RouteComponentScoped``` like that one.
But of course there are some downside of this concept. 
- Have to use the qualifier on injection point, and on the bean class too. like this:
```
@RouteComponentScoped
@RouteComponentGroup(MyRouteLayout.class)
public class MyScopedComponent { ... 
```
 ```@Inject @RouteComponentGroup(MyRouteLayout.class) MyScopedComponent component;```
- Groups are always open, because the scope would be activate by ( and would store to ) the UI scope.
- A bean belongs to one group. So you can't define this way a bean for a lifetime of any view. But we can't really define what is a view, so it is affordable :) Maybe we need an other scope, or scopes too for this purpose. Like a scope lives until the next navigation, or a scope lives until the next route target change.

It is easy to add a detach listener destroying the group of the route component. But can't guarantee this will be the last listener. 
It is questionable when should the detach happen. Issues like detach after context close would not encounter even with the current behaviour.

denis-anisimov:
fixes #263

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
@@ -16,9 +16,9 @@
 package com.vaadin.flow.spring;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.URL;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -29,13 +29,12 @@ import com.vaadin.flow.di.Instantiator;
 import com.vaadin.flow.function.DeploymentConfiguration;
 import com.vaadin.flow.server.ServiceException;
 import com.vaadin.flow.server.SessionDestroyListener;
+import com.vaadin.flow.server.UIInitListener;
 import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.server.VaadinServlet;
 import com.vaadin.flow.server.VaadinServletService;
 import com.vaadin.flow.server.VaadinSession;
-import com.vaadin.flow.server.WebBrowser;
 import com.vaadin.flow.shared.Registration;
-import com.vaadin.flow.theme.AbstractTheme;
 
 /**
  * Spring application context aware Vaadin servlet service implementation.
@@ -95,6 +94,14 @@ public class SpringVaadinServletService extends VaadinServletService {
         }
         return spiInstantiator.isPresent() ? spiInstantiator
                 : springInstantiators.stream().findFirst();
+    }
+
+    @Override
+    public void init() throws ServiceException {
+        super.init();
+        Map<String, UIInitListener> uiInitListeners = context
+                .getBeansOfType(UIInitListener.class);
+        uiInitListeners.values().forEach(this::addUIInitListener);
     }
 
     @Override
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.spring.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.beans.factory.annotation.Qualifier;
+
+import com.vaadin.flow.component.HasElement;
+import com.vaadin.flow.router.HasErrorParameter;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouterLayout;
+
+/**
+ * TODO : docs
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD,
+        ElementType.PARAMETER })
+public @interface RouteScopeOwner {
+
+    /**
+     * Owner class of the qualified {@link RouteScope @RouteScope} bean.
+     * <p>
+     * A {@link Route @Route}, or a {@link RouterLayout}, or a
+     * {@link HasErrorParameter}
+     *
+     * @return owner class
+     */
+    Class<? extends HasElement> value();
+}
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
@@ -28,7 +28,10 @@ import com.vaadin.flow.router.Route;
 import com.vaadin.flow.router.RouterLayout;
 
 /**
- * TODO : docs
+ * Link a {@link RouteScope @RouteScope} bean to its owner.
+ * <p>
+ * Owner is a router component. A {@link Route @Route}, or a
+ * {@link RouterLayout}, or a {@link HasErrorParameter}.
  * 
  * @author Vaadin Ltd
  * @since
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
@@ -104,27 +104,39 @@ class BeanStore implements Serializable {
 
     Void doDestroy() {
         assert session.hasLock();
-        for (Runnable destructionCallback : destructionCallbacks.values()) {
-            try {
-                destructionCallback.run();
-            } catch (Exception e) {
-                LOGGER.error("BeanStore destruction callback failed", e);
+        try {
+            for (Runnable destructionCallback : destructionCallbacks.values()) {
+                try {
+                    destructionCallback.run();
+                } catch (Throwable throwable) {
+                    LOGGER.error("BeanStore destruction callback failed",
+                            throwable);
+                }
             }
+        } finally {
+            destructionCallbacks.clear();
+            objects.clear();
         }
-        destructionCallbacks.clear();
-        objects.clear();
         return null;
     }
 
     protected Object doRemove(String name) {
-        destructionCallbacks.remove(name);
+        Runnable destructionCallback = destructionCallbacks.remove(name);
+        try {
+            if (destructionCallback != null) {
+                destructionCallback.run();
+            }
+        } catch (Throwable throwable) {
+            LOGGER.error("BeanStore destruction callback failed", throwable);
+        }
         return objects.remove(name);
     }
 
     protected Object doGet(String name, ObjectFactory<?> objectFactory) {
         Object bean = objects.get(name);
         if (bean == null) {
-            storeBean(name, objectFactory.getObject());
+            bean = objectFactory.getObject();
+            storeBean(name, bean);
         }
         return bean;
     }
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
@@ -116,18 +116,21 @@ class BeanStore implements Serializable {
         return null;
     }
 
-    private Object doRemove(String name) {
+    protected Object doRemove(String name) {
         destructionCallbacks.remove(name);
         return objects.remove(name);
     }
 
-    private Object doGet(String name, ObjectFactory<?> objectFactory) {
+    protected Object doGet(String name, ObjectFactory<?> objectFactory) {
         Object bean = objects.get(name);
         if (bean == null) {
-            bean = objectFactory.getObject();
-            objects.put(name, bean);
+            storeBean(name, objectFactory.getObject());
         }
         return bean;
+    }
+
+    protected void storeBean(String name, Object bean) {
+        objects.put(name, bean);
     }
 
     private <T> T execute(Supplier<T> supplier) {
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
@@ -120,7 +120,7 @@ class BeanStore implements Serializable {
         return null;
     }
 
-    protected Object doRemove(String name) {
+    Object doRemove(String name) {
         Runnable destructionCallback = destructionCallbacks.remove(name);
         try {
             if (destructionCallback != null) {
@@ -132,7 +132,7 @@ class BeanStore implements Serializable {
         return objects.remove(name);
     }
 
-    protected Object doGet(String name, ObjectFactory<?> objectFactory) {
+    Object doGet(String name, ObjectFactory<?> objectFactory) {
         Object bean = objects.get(name);
         if (bean == null) {
             bean = objectFactory.getObject();
@@ -141,7 +141,7 @@ class BeanStore implements Serializable {
         return bean;
     }
 
-    protected void storeBean(String name, Object bean) {
+    void storeBean(String name, Object bean) {
         objects.put(name, bean);
     }
 
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
@@ -1,0 +1,311 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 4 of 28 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
```diff
@@ -104,32 +104,47 @@ VaadinSession getVaadinSession() {
...
         Object bean = objects.get(name);
         if (bean == null) {
             bean = objectFactory.getObject();
-            objects.put(name, bean);
+            storeBean(name, bean);
         }
         return bean;
     }
 
+    protected void storeBean(String name, Object bean) {
```
pleku, line 144: Protected method without javadocs

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -0,0 +1,380 @@
...
+                routeStores.put(key, beanStore);
+            }
+            return beanStore;
+        }
+
+        private void relocateStore(UI ui, String key) {
+            assert session.hasLock();
+            BeanStore beanStore = routeStores.remove(key);
+            if (beanStore == null) {
+                // TODO
```
pleku, line 111: ??? By minimum would need commenting on _what_ is to be done

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -0,0 +1,380 @@
...
+ * @see com.vaadin.flow.spring.annotation.VaadinSessionScope
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+public class VaadinRouteScope extends AbstractScope implements UIInitListener {
+
+    public static final String VAADIN_ROUTE_SCOPE_NAME = "vaadin-route";
+
+    private static class RouteStoreWrapper implements Serializable {
```
pleku, line 68: Non-blocking: I think `SessionRouteStore` or such would have been a more descriptive name than just _wrapper_

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -0,0 +1,380 @@
...
+
+    @Override
+    public String getConversationId() {
+        return getVaadinSession().getSession().getId() + "-route-scope";
+    }
+
+    @Override
+    public void uiInit(UIInitEvent event) {
+        NavigationListener listener = new NavigationListener(event.getUI());
+        ComponentUtil.setData(event.getUI(), NavigationListener.class,
```
pleku, line 330: Should the data be cleared at some point to reduce the risk that detached UIs would remain referenced from the VaadinRouteScope or RouteBeanStore ? I don't think it is referenced (and thus leaked) in this PR, but there is a risk that some future change could start leaking memory (at least in theory).

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -0,0 +1,380 @@
...
+        }
+
+        private BeanStore getBeanStore(UI ui) {
+            assert session.hasLock();
+            ExtendedClientDetails details = ui.getInternals()
+                    .getExtendedClientDetails();
+            String key = getUIStoreKey(ui);
+            if (details == null) {
+                ui.getPage().retrieveExtendedClientDetails(
+                        det -> relocateStore(ui, key));
```
pleku, line 97: If the UI is detached before this callback is run, will the UI be garbage collected ?

pleku (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
```diff
@@ -0,0 +1,55 @@
...
+ * Owner is a router component. A {@link Route @Route}, or a
+ * {@link RouterLayout}, or a {@link HasErrorParameter}.
+ * 
+ * @author Vaadin Ltd
+ * @since
+ *
+ */
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD,
```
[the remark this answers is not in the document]
denis-anisimov: https://github.com/vaadin/spring/blob/master/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinApplicationConfiguration.java#L47

In CDI this is called a producer.
You may make a bean using a method. The method which produces a bean may have a scope.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
```diff
@@ -104,32 +104,47 @@ VaadinSession getVaadinSession() {
...
+            objects.clear();
         }
-        destructionCallbacks.clear();
-        objects.clear();
         return null;
     }
 
-    private Object doRemove(String name) {
-        destructionCallbacks.remove(name);
+    protected Object doRemove(String name) {
```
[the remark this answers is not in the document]
denis-anisimov: `class BeanStore`

The class is effectively private: at least internal.
This class is never supposed to be opened for the users.
`protected` in fact makes it more opened than package local but it explicitely makes a method to be used by subclasses and not by another classes inside the package.

But OK, I can make it package local.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
```diff
@@ -104,32 +104,47 @@ VaadinSession getVaadinSession() {
...
+                destructionCallback.run();
+            }
+        } catch (Throwable throwable) {
+            LOGGER.error("BeanStore destruction callback failed", throwable);
+        }
         return objects.remove(name);
     }
 
-    private Object doGet(String name, ObjectFactory<?> objectFactory) {
+    protected Object doGet(String name, ObjectFactory<?> objectFactory) {
```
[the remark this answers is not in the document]
denis-anisimov: Made package local

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/BeanStore.java
denis-anisimov: package local

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
denis-anisimov: Leftover. Thanks.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
denis-anisimov: Every "store" is stored inside a session. So `SessionRouteStore` name doesn't bring a new information.

`RouteStoreWrapper` is not a best name but I can't suggest any better and I don't see `SessionRouteStore`  as a better name.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
denis-anisimov: I don't understand this comment.

It looks like it's attached to `ComponentUtil.setData(event.getUI(), NavigationListener.class,` line.
This line stores a navigation listener which is 1:1 to `UI` into the `UI`.
This is a cyclic reference which will be properly GCed.

I don't see how are `VaadinRouteScope` and `RouteBeanStore` applicable here.

denis-anisimov: `VaadinRouteScope` is a singleton bean and it doesn't store anything inside itself.
`RouteBeanStore` is 1:1 to UI and also will be properly GCed.

The only valid point is `RouteStoreWrapper` may keep extra data per `UI` which can be detached.
But: 
- it's data without the `UI` itself
- it's stored per session and will be removed on session remove.
- it's not a place to care about `RouteStoreWrapper` .

But I've fixed this case.

vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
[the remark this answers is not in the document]
pleku: I know that you can do that - just wondering if it makes sense with routescope+routescopeowner. I guess one utilize it to customize the bean that will be injected

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
denis-anisimov: `UI` is not stored anywhere.
Its ID or window name is used to store a data.

This is a valid comment for data stored per detached `UI` and this is fixed now.

vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScopeOwner.java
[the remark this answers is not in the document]
denis-anisimov: Why not ?

It's just a way of instantiating the class.
Normally you don't care as class is instituted: it's delegated to Spring via CTOR + may be autowired CTOR parameters.
If for some reason I want to instantiate a class by myself then I will do a producer.
Then it's just a question which scope I use. Don't see any difference with or without scope.
I don't know why would I want to do this but this is absolutely fine to do this with the route scope or any other scope.

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
```diff
@@ -97,6 +96,14 @@ public SpringVaadinServletService(VaadinServlet servlet,
                 : springInstantiators.stream().findFirst();
     }
 
+    @Override
+    public void init() throws ServiceException {
+        super.init();
+        Map<String, UIInitListener> uiInitListeners = context
+                .getBeansOfType(UIInitListener.class);
+        uiInitListeners.values().forEach(this::addUIInitListener);
```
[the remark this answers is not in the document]
denis-anisimov: Done.

pleku (approved):

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
pleku: I disagree, it would add more value as routes are generally registered to application scope and secondarily to session scope, even while here for routescope implementation, it only makes sense to store beans to session. I dislike "wrapper" naming because it there is typically reason why things are wrapped, and here is not even a wrapper but a holder of the route stores per specific session. Thus it is like `RouteStoreStore` / or `SessionRouteStoreMap` orwhatnot. But like I said, it was not blocking.

PR #799, opened 2021-05-17
test: add fusion form integration tests

haijian-vaadin:
Depends on https://github.com/vaadin/flow/pull/10968


platosha (approved):

vaadin-spring-tests/test-ts-form/frontend/views/form/vaadin-elements-view.ts
```diff
@@ -0,0 +1,113 @@
...
+import '@vaadin/vaadin-button';
+import '@vaadin/vaadin-notification';
+
+import { field, Binder } from '@vaadin/form';
+
+import ElementsModel from 'Frontend/generated/com/vaadin/flow/spring/form/data/ElementsModel';
+import { ElementsEndpoint } from 'Frontend/generated/ElementsEndpoint';
+
+@customElement('vaadin-elements-view')
+export class VaadinElementsView extends LitElement {
```
Haprog, line 34: Since this PR also includes the View and Layout classes from Vaadin Start this should probably extend View instead of LitElement directly. Then we can also leave out the `createRenderRoot()` override at the end. Or is there some reason to include those helpers from Vaadin Start but not use them?

Haprog (changes requested):

vaadin-spring-tests/test-ts-form/frontend/views/form/vaadin-elements-view.ts
haijian-vaadin: good point.
Also realized that we can go one step further, we can reuse the existing fusion security test module for form as well instead of creating yet another module.

Haprog (approved):

PR #800, opened 2021-05-18
feat: RouteScope without an owner means maps scope to the current Route

### LINKED ISSUE #798: RouteScope: change the implementation to allow having a bean without RouteScopeOwner so that in that case the owner is the current route target
Currently it's possible to use `RouteScope`  without `RouteScopeOwner`.
In this case the semantic is the same as a prototype scope: the bean is never stored anywhere and is not reused.
As a result it has no sense to use it this way since there is already prototype scope.

The suggestion is :
change the impl so that the absence of the `RouteScopeOwner` means the scope of the current route target.
It allows to simplify  `RouteScope` without RouteScopeOwner` usage in mpr .

Also it will allow to keep components inside the route target which are not directly referenced : they will be reused (e.g. a toggle button which replaces the content from one component to another staying inside the same route target).

denis-anisimov:
fixes #798

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/routescope/BeansWithNoWOnerView.java
+++ b/vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/routescope/BeansWithNoWOnerView.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.spring.test.routescope;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+
+import com.vaadin.flow.component.AttachEvent;
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouterLink;
+
+@Route("beans-no-owner")
+public class BeansWithNoWOnerView extends Div {
+
+    @Autowired
+    private ApplicationContext context;
+
+    private boolean isSubDiv;
+
+    private Component current;
+
+    @Override
+    protected void onAttach(AttachEvent attachEvent) {
+        NativeButton button = new NativeButton("switch content", ev -> {
+            remove(current);
+            if (isSubDiv) {
+                current = context.getBean(ButtonNoOwner.class);
+            } else {
+                current = context.getBean(DivNoOwner.class);
+            }
+            add(current);
+            isSubDiv = !isSubDiv;
+        });
+        button.setId("switch-content");
+        add(button);
+
+        RouterLink link = new RouterLink("another-view",
+                AnotherBeanNopOwnerView.class);
+        link.getElement().getStyle().set("display", "block");
+        link.setId("navigate-another");
+        add(link);
+
+        current = context.getBean(ButtonNoOwner.class);
+        add(current);
+    }
+
+}
--- a/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/routescope/BeansWithNoOWnerIT.java
+++ b/vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/routescope/BeansWithNoOWnerIT.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.spring.test.routescope;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+import com.vaadin.flow.spring.test.AbstractSpringTest;
+
+public class BeansWithNoOWnerIT extends AbstractSpringTest {
+
+    @Override
+    protected String getTestPath() {
+        return "/beans-no-owner";
+    }
+
+    @Test
+    public void beansWithNoOwner_preservedWithinTheSameRouteTarget_notPreservedAfterNavigation() {
+        open();
+
+        WebElement button = findElement(By.id("no-owner-button"));
+        String buttonId = button.getText();
+
+        switchContent();
+
+        WebElement div = findElement(By.id("no-owner-div"));
+        String divId = div.getText();
+
+        switchContent();
+
+        Assert.assertEquals("Button component is not preserved", buttonId,
+                findElement(By.id("no-owner-button")).getText());
+
+        switchContent();
+
+        Assert.assertEquals("Div component is not preserved", divId,
+                findElement(By.id("no-owner-div")).getText());
+
+        navigateTo("navigate-another");
+
+        Assert.assertNotEquals(
+                "Div with RouteScope and no owner is preserved during "
+                        + "navigation but it should not be preserved",
+                divId, findElement(By.id("another-no-owner")).getText());
+
+        navigateTo("no-owner-view");
+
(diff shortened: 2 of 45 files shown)


vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/routescope/BeansWithNoWOnerView.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.vaadin.flow.component.AttachEvent;
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouterLink;
+
+@Route("beans-no-owner")
+public class BeansWithNoWOnerView extends Div {
```
pleku, line 29: Not-blocking: typo _WOner_

vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/routescope/BeansWithNoOWnerIT.java
```diff
@@ -0,0 +1,78 @@
...
+    public void beansWithNoOwner_preservedWithinTheSameRouteTarget_notPreservedAfterNavigation() {
+        open();
+
+        WebElement button = findElement(By.id("no-owner-button"));
+        String buttonId = button.getText();
+
+        switchContent();
+
+        WebElement div = findElement(By.id("no-owner-div"));
+        String divId = div.getText();
```
pleku, line 42: Non-blocking: it is slightly confusing to name this as _id_ as it as a different meaning in this context (element id), using _uuid_ would have been clearer.

vaadin-spring/src/main/java/com/vaadin/flow/spring/annotation/RouteScope.java
```diff
@@ -29,20 +29,20 @@
...
- * remain in the scope.
+ * As long as the owner component stays attached, all beans owned by it remain
+ * in the scope.
  * <p>
- * When a scoped bean is a router component, an owner can be any ancestor
- * {@link RouterLayout}, or the bean itself. Without the {@link RouteScopeOwner}
- * annotation the scope effectively behaves as prototype scope.
+ * Without the {@link RouteScopeOwner} annotation the owner is the current route
+ * target (so it's dynamically calculated). In this case the beans are preserved
+ * as long as the navigation target stays the same.
```
pleku, line 45:
```suggestion
 * Without the {@link RouteScopeOwner} annotation the owner is the current route
 * target component (dynamically calculated). With nested routing hierarchies, the
 * target is the "leaf" or "bottom most" routing component. The beans are
 * preserved as long as the navigation target stays the same.
```

pleku (approved):
One suggestion for javadocs and two non blocking comments

vaadin-spring-tests/test-spring-common/src/main/java/com/vaadin/flow/spring/test/routescope/BeansWithNoWOnerView.java
denis-anisimov: done

vaadin-spring-tests/test-spring-common/src/test/java/com/vaadin/flow/spring/test/routescope/BeansWithNoOWnerIT.java
denis-anisimov: It's an id for the div abstract component.
There is an element `id` , yes but it's not used anyhow here.
It's a bit confusing, yes, but I can't find a better name here.

pleku (changes requested):
One weird case that came to my mind:

In case you inject a bean
`@RouteScope class FooBean {} `
to
`@Route("parent") class ParentView implements RouteLayout {}`
, and then navigate to a subview of `ParenView` 
`@Route(value="child", layout=ParentView.class) class ChildView {}`
and inject `FooBean` there,
will the injected bean be a new one or the same as in `ParentView` ?
My first impression would be that yes - based on the annotation javadocs as the `ParentLayout` is still attached. But I think that impression is wrong since the injected component will be in new "context" in the subview and live as long as that is attached. Can we have a test for this ?

denis-anisimov:
I don't understand the question.

>`@Route("parent") class ParentView implements RouteLayout {}`

This is a declaration .
Does it have injection of `FooBean`? I don't see this from the descrption.


>, and then navigate to a subview of ParentView

Then ?
Was there another navigation target where the navigation is at the moment ?

>`@Route(value="child", layout=ParentView.class) class ChildView {}`

Navigation happens here: once. `ParentView` never used before navigation happens into the `ChildView`.

Once navigation happens the injected bean into `ParentView`  (if any) will have scope of the current bottom-most `ChildView` navigation target.

So both injections will have the same instance : yes: because the  `ChildView` defines the scope.

Then if there is _another_ navigation target with the same parent `ParentView` then the injection in the parent will stay the same (because the reference to the bean is hard and may not be changed automagically) but the new navigation target will get a new instance of the bean.
And this is another example of wrong scope usage.

So what do you suggest to test exactly ?

pleku:
> I don't understand the question.
> 
> > `@Route("parent") class ParentView implements RouteLayout {}`
> 
> This is a declaration .
> Does it have injection of `FooBean`? I don't see this from the descrption.

The description says that you inject the bean to `ParentView` 
_In case you inject a bean ... to ..._

> > , and then navigate to a subview of ParentView
> 
> Then ?
> Was there another navigation target where the navigation is at the moment ?

Yes, sorry that I wasn't explicit as I thought it would be obvious that the `ParentView` is opened first and then the navigation to `ChildView` happens ...

> > `@Route(value="child", layout=ParentView.class) class ChildView {}`
> 
> Navigation happens here: once. `ParentView` never used before navigation happens into the `ChildView`.
> 
> Once navigation happens the injected bean into `ParentView` (if any) will have scope of the current bottom-most `ChildView` navigation target.
> 
> So both injections will have the same instance : yes: because the `ChildView` defines the scope.
> 

This is obvious. And not what I meant - the question was about having the parent first open.

> Then if there is _another_ navigation target with the same parent `ParentView` then the injection in the parent will stay the same (because the reference to the bean is hard and may not be changed automagically) but the new navigation target will get a new instance of the bean.
> And this is another example of wrong scope usage.
> 
> So what do you suggest to test exactly ?

Given the code (maybe compiles) 
```java
@SpringComponent
@RouteScope
public class FooBean {

    private String uuid = UUID.randomUUID().toString();

    public String getText() {
        return uuid;
    }
}

@Route("parent")
public class ParentView extends Div implements RouterLayout {

    public ParentView (@Autowired FooBean bean) {
        add(new Span("UUID; "+bean.getText()), new RouterLink("Open child",  ChildView.class));
    }
}

@Route(value = "child", layout = ParentView.class)
public class ChildView extends Div {

    public ChildView(@Autowired FooBean bean) {
        setText("UUID: "+bean.getText());
    }
}
```
1) First open /parent 
2) then click navigation to child view.
Will both get the same bean or not ? This case is not clear to me 

If the page is refreshed - then they will get the same bean that is owned by `ChildView` like expected.

denis-anisimov:
OK, now I see a confusing expectation.

Even though the javadocs says:
`The beans are preserved as long as the navigation target stays the same.`
So the _target stays the same_ (and not "object is preserved").

This is covered in the docs but I agree that there is always a way to misinterpret what docs say and there will be always 
someone who thinks it should work in the way which it's not intended to work and we will have definitely issues about that.

This is again a magic since there is no clear contract expressed via Java type safe way. And that's why I don't like the owner absence : it gives too many ways to invent your own way of how it should work which doesn't work and then complain about this.

I will add the test even though technically there is already such test though it doesn't preserve the hierarchy.
Effectively this is just `@RouteSopeOwner(ParentView.class)` for the first navigation and it's `@RouteSopeOwner(ChildView.class)`  for the second navigation .

denis-anisimov:
Alright.............
Now it's a confusing expectation for me.

In fact the bean is _preserved_.
This is because the implementation for explicit owner is reused in this case but this case in fact should be handled specifically
if we _want to work it so that it's not preserved_.

Now the question is:
- should I keep the current simple implementation which follows the genetic case and then update the javadocs so that it says "until the component in the navigation chain is preserved" instead of "route target stays the same.
- or should I reimplement this so that the bean is not preserved ? That will allow to keep the javadocs as is. That will require some efforts since this is a very specific case.

Technically I'm fine with both of these approaches. The first one makes sense and the only confusion is introduced by the docs. It's aligned with the generic case (since during the first navigation the bean got its owner `ParentView.class` and the parent is not lost).

The second also makes sense but that will complicate the code and will require special case maintenance.

So the only question : which way it should be?

pleku:
Thinking in terms of the usage - to me `RouteScope` without explicit owner means that it is injected to the _currently active navigation target component scope_, like it is done now. And as long as that component stays _active_ (to me this means the component is attached) the bean should be the same. So I would go with the first case but clarify this explicitly in the javadocs and have an IT that validates that this is by design and nobody will break it accidentally without removing the test (which would be more visible in review).

For the second case - I would just stay that people should use the prototype scope instead of `RouteScope`. They will not get the predestroy lifecycle method working, but they can always use the detach method of the component instead if necessary.

denis-anisimov:
OK.

denis-anisimov:
Done

pleku (approved):

PR #806, opened 2021-05-21
test: Run Flow security tests also with a context path

Artur-:


taefi (approved):
While "SSH" would not be my choice for shortening "Spring Security Helper", LGTM! ;-)

PR #815, opened 2021-05-24
fix: annotate default error handler

### LINKED ISSUE #661: When deploying to a Tomcat server, SpringRouteNotFoundError is still active although being fixed
Related issue: https://github.com/vaadin/flow/issues/8948

I checked the war and webapps, and all of them contain **14.0.2**. When locally tested, it works, but when deployed to my webserver **it still shows the error of the related issue**.

```
Version: 14.0.2
Vaadin Version: 17.0.3
Java: Adopt Openjdk 14
```

caalador:
Annotate SpringRouteNotFoundError with
DefaultErrorHandler so that users can
make custom error views for NotFoundException

requires vaadin/flow#11063

fixes #661


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -355,28 +355,12 @@ public void failFastContextInitialized(ServletContextEvent event) {
...
-                        clazz, HasErrorParameter.class);
-                return exceptionType != null && exceptionType
-                        .isAssignableFrom(NotFoundException.class);
-            })) {
-                errorComponents.add(SpringRouteNotFoundError.class);
-            }
-            registry.setErrorNavigationTargets(errorComponents);
+                            .map(clazz -> (Class<? extends Component>) clazz);
+            registry.setErrorNavigationTargets(
+                    hasErrorComponents.collect(Collectors.toSet()));
```
pleku, line 363: What code is actually deciding that `SpringRouteNotFoundError` will be used instead of `RouteNotFoundError` ? Since both have `@DefaultErrorHandler`.

pleku (changes requested):
One question and I wonder if an IT should be added too ...

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
caalador: The choosing code is the default selection as SpringRouteNotFoundError is picked as it extends the RouteNotFoundError so the DefaultHandler annotation is not checked in that case.

pleku (approved):
Will merge-from-master after #820 is merged

PR #829, opened 2021-05-27
Destroy beans after navigation out of the route and refresh

### LINKED ISSUE #818: @PreDestroy method of @RouteScope bean is not invoked after page refresh
### Description of the bug / feature
Suppose the configuration with `@RouteScope` annotated bean being auto-wired into `@PreserveOnRefresh` annotated route component:

```java
@RouteScope
@SpringComponent
public class MessageService {

    @PostConstruct
    public void init() {
        System.out.println("Initialized");
    }

    @PreDestroy
    public void cleanUp() {
        System.out.println("Cleaned up");
    }
}

@PreserveOnRefresh
@Route(value = "message")
public class MessageView extends VerticalLayout {

    private final MessageService messageService;

    public MessageView(@Autowired MessageService messageService) {
        this.messageService = messageService;
        add(new Button("Unsubscribe", click ->
                UI.getCurrent().navigate(MainView.class)));
    }
}

@Route("")
public class MainView extends VerticalLayout {
    public MainView() {
        Button navigateToMessages = new Button("Subscribe",
                event -> UI.getCurrent().navigate(MessageView.class));
        add(navigateToMessages);
    }
}
```
`MessageService` gets out-of-scope when navigating to main view and its `@PreDestroy` method is invoked as expected.
But as soon as one refreshes the message view page, the message bean then "extends" its scope and is preserved between main view <-> message view navigations. Thus, `@PreDestroy` method is not invoked anymore.

### Minimal reproducible example
[skeleton-starter-flow-spring-pre-destroy.zip](https://github.com/vaadin/spring/files/6540314/skeleton-starter-flow-spring-pre-destroy.zip)
1. Open and start the attached project
2. Go to http://localhost:8080
3. Click _Subscribe_ button
4. Notice 'Initialized' message in the app console
5. Click _Unsubscribe_ button
6. Notice 'Cleaned up' message in the app console
7. Click _Subscribe_ button again
8. Refresh the browser page
9. Click _Unsubscribe_ button
10. Notice no 'Cleaned up' message present in the app console, and the same bean reference is used.

### Expected behavior
`@PreDestroy` method always invoked
### Actual behavior
`@PreDestroy` method is not invoked after page refresh
### Versions:
- Vaadin / Flow version/ Spring add-on version: 21.0.0.alpha2 / 8.0.0.alpha3 / 18.0.0.alpha1
- Java version: 15
- OS version: Mac OS
- Browser version (if applicable): Chrome 90
- Application Server (if applicable): spring-boot
- IDE (if applicable): IntelliJ IDEA Ultimate

denis-anisimov:
fixes #818

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
@@ -140,6 +140,8 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
             session.lock();
             try {
                 session.setAttribute(RouteStoreWrapper.class, null);
+                routeStores.values().forEach(BeanStore::destroy);
+                routeStores.clear();
             } finally {
                 session.unlock();
                 if (sessionDestroyListenerRegistration != null) {
@@ -308,7 +310,7 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
         private boolean resetUI() {
             for (UI ui : getVaadinSession().getUIs()) {
                 String windowName = getWindowName(ui);
-                if (ui != currentUI
+                if (ui != currentUI && windowName != null
                         && windowName.equals(getWindowName(currentUI))) {
                     currentUI = ui;
                     return true;
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.spring.scopes;
 import javax.servlet.ServletContext;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -72,7 +73,7 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
 
         private final Registration sessionDestroyListenerRegistration;
 
-        private final Map<String, BeanStore> routeStores;
+        private final Map<String, RouteBeanStore> routeStores;
 
         private RouteStoreWrapper(VaadinSession session) {
             assert session.hasLock();
@@ -88,7 +89,7 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
             }
         }
 
-        private BeanStore getBeanStore(UI ui) {
+        private RouteBeanStore getBeanStore(UI ui) {
             assert session.hasLock();
             ExtendedClientDetails details = ui.getInternals()
                     .getExtendedClientDetails();
@@ -97,7 +98,7 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
                 ui.getPage().retrieveExtendedClientDetails(
                         det -> relocateStore(ui, key));
             }
-            BeanStore beanStore = routeStores.get(key);
+            RouteBeanStore beanStore = routeStores.get(key);
             if (beanStore == null) {
                 beanStore = new RouteBeanStore(ui, session,
                         uiInstance -> routeStores
@@ -109,7 +110,7 @@ public class VaadinRouteScope extends AbstractScope implements UIInitListener {
 
         private void relocateStore(UI ui, String key) {
             assert session.hasLock();
-            BeanStore beanStore = routeStores.remove(key);
+            RouteBeanStore beanStore = routeStores.remove(key);
             if (beanStore == null) {
                 LoggerFactory.getLogger(RouteStoreWrapper.class).trace(
                         "UI bean store is not found by the initial UI id via the key '"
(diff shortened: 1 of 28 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -398,4 +434,16 @@ private static UI getUI() {
         return ui;
     }
 
+    private static UI findPreservingUI(UI ui) {
+        VaadinSession session = ui.getSession();
+        for (UI sessionUi : session.getUIs()) {
+            String windowName = getWindowName(ui);
```
mshabarov, line 440: Should it be `String windowName = getWindowName(sessionUi);` instead?
Otherwise `windowName.equals(getWindowName(ui))` compares the same things.

denis-anisimov: Good catch.
Leftover after refactoring.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
```diff
@@ -159,31 +160,32 @@ private void destroy() {
         private Class<?> currentNavigationTarget;
         private List<Class<? extends RouterLayout>> currentLayouts;
 
-        private Map<Class<?>, Set<String>> beanNamesByNavigationComponents = new HashMap<>();
```
mshabarov, line 162: If I recall the root cause properly, the bug was in preserving the route scoped bean even outside of owner navigation target, when refresh is done in the browser. And that's why the destroy method was not invoked, because the bean itself was not destroyed. From a very 'outer' point of view, something was broken after refresh event, making the beans 'extend' their scope.

From the changes in the current PR, I'm not sure what is the essential point the changed code, which fixes that. 
I clearly see that the `beanNamesByNavigationComponents` map/holder is now moved to `RouteBeanStore` object, which is in turn mapped to UI objects and stored in the user's session.

So, do I understand right, that before this change the UI object was destroyed after refresh and this map was lost. And **this was the root cause of preserving beans**? Now this map is in the session, so browser refresh does not impact it.

denis-anisimov: yes.

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
mshabarov: Tests pass even with `String windowName = getWindowName(ui);`. Is it possible to cover it with extra test, or would it take efforts/mocking?

vaadin-spring/src/test/java/com/vaadin/flow/spring/scopes/VaadinRouteScopeTest.java
```diff
@@ -150,4 +125,102 @@ public void destroySession_sessionAttributeIsCleanedAndDestructionCallbackIsCall
...
+        UI.setCurrent(anotherUI);
+
+        scope = initScope(anotherUI);
+
+        // the bean is not removed since there is a "preserved" UI
+        Assert.assertEquals(0, count.get());
+
+        navigateTo(anotherUI, new AnotherNavigationTarget());
+
+        // the bean is not removed since there is a "preserved" UI
```
mshabarov, line 168:
```suggestion
        // the bean is removed since navigation away from it's owner navigation target
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/scopes/VaadinRouteScopeTest.java
```diff
@@ -150,4 +125,102 @@ public void destroySession_sessionAttributeIsCleanedAndDestructionCallbackIsCall
...
+        UI anotherUI = makeAnotherUI(ui);
+
+        ExtendedClientDetails details = Mockito
+                .mock(ExtendedClientDetails.class);
+        Mockito.when(details.getWindowName()).thenReturn("bar");
+        ui.getInternals().setExtendedClientDetails(details);
+        anotherUI.getInternals().setExtendedClientDetails(details);
+
+        ui.getSession().addUI(ui);
+        ui.getSession().addUI(anotherUI);
```
mshabarov, line 141: Add one more UI with another `windowName`. It would check that current UI is replaced by the same window name, and not with another window name, i.e. covers this `windowName.equals(getWindowName(currentUI))`

mshabarov (changes requested):

vaadin-spring/src/test/java/com/vaadin/flow/spring/scopes/VaadinRouteScopeTest.java
denis-anisimov: another test is added for this case

vaadin-spring/src/main/java/com/vaadin/flow/spring/scopes/VaadinRouteScope.java
denis-anisimov: the test is added

vaadin-spring/src/test/java/com/vaadin/flow/spring/scopes/VaadinRouteScopeTest.java
```diff
@@ -150,4 +125,149 @@ public void destroySession_sessionAttributeIsCleanedAndDestructionCallbackIsCall
...
+        count.set(0);
+
+        UI.setCurrent(anotherUI);
+
+        scope = initScope(anotherUI);
+
+        navigateTo(anotherUI, new AnotherNavigationTarget());
+
+        // the bean is not removed since there is a "preserved" UI
+        Assert.assertEquals(0, count.get());
```
mshabarov, line 216: Hmm, why is no pre-destroy calls expected here? As anotherUI has no details and no window name, I would expect bean is destroyed, rather than UI is reset.

denis-anisimov: It's worth to check several line above where `Assert.assertEquals(1, count.get());` is written.

mshabarov (approved):

PR #831, opened 2021-06-02
chore: use smaller images

haijian-vaadin:
use smaller images to see if it could the security tests on CI more stable


haijian-vaadin:
It doesn't help, so close this one

PR #849, opened 2021-06-17
test: Test uploads when using Spring Security

Artur-:


Legioth:
Shouldn't we instead change it so that `requestUtil::isFrameworkInternalRequest` also detects upload requests, rather than doing something specific to Spring Security?

Artur-:
Maybe it could be considered an "internal request that is needed for the application to work"

Artur-:
Depends on https://github.com/vaadin/flow/pull/11278 then

pleku (approved):

pleku:
Will rerun the build once the flow snapshot build has finished.

PR #873, opened 2021-08-11
Launch a browser automatically when starting an app in development mode

### LINKED ISSUE #786: Integrate LaunchUtil from a-vaadin-helper
The LaunchUtil helper in https://github.com/Artur-/a-vaadin-helper/blob/master/src/main/java/org/vaadin/artur/helpers/LaunchUtil.java is used in all new projects from start.vaadin.com to launch a browser instance when the application has started. It should be integrated into this project

Artur-:
## Description

Integrates the launch util from https://github.com/Artur-/a-vaadin-helper/blob/master/src/main/java/org/vaadin/artur/helpers/LaunchUtil.java

Fixes #786

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

- [] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


Artur-:
Ideas on how to auto test this? If you run in headless mode as the build does, you cannot open a browser

vaadin-spring/src/main/java/com/vaadin/flow/spring/LaunchUtil.java
```diff
@@ -0,0 +1,137 @@
...
+    public static void launchBrowserInDevelopmentMode(
+            ConfigurableApplicationContext app) {
+        if (isLaunched()) {
+            // Only launch browser on startup, not on reload
+            return;
+        }
+        if (!LaunchUtil.isProductionMode(app)) {
+            String location = "http://localhost:"
+                    + app.getEnvironment().getProperty("local.server.port")
+                    + "/";
```
knoobie, line 51: Should this open always on "/" or would it be better to add the context path as well? ServletContext should be available

Artur-: Good catch. This was never supported by the helper library but now it supports both context path and Vaadin mapping

pleku (changes requested):
My biggest issue is with the fact that I don't understand _why_ we would want to include this to the `vaadin-spring` artifact. I don't think this has anything to do with the integration of the two or building applications with Vaadin & Spring.

In my opinion this should be in some _yet another_ module related to vaadin-spring development time, if not so that is could be in `vaadin-dev-server` so that it only works when spring is detected to be used (optional dependencies).
EDIT: since it depends on spring. it would probably have to be in this project so that it is built against the same version. 

And additionally I am also concerned for the fact that we don't really even want to maintain this as a part of our codebase, as we have enough on our plate anyway.

Artur-:
Doesn't sound fruitful to add something like `vaadin-spring-dev` just for this. If there are more dev time features, then it could make sense. As you noted, `vaadin-dev-server` won't work unless we are fine to add a spring dependency there

kolle1986:
I'm using latest version of a-vaadin-helper 1.7.2 but it still doesn't use server.servlet.context-path for path url in launchBrowserInDevelopmentMode. always launcing empty url (just localhost:8080)

pleku:
> Doesn't sound fruitful to add something like vaadin-spring-dev just for this. If there are more dev time features, then it could make sense. As you noted, vaadin-dev-server won't work unless we are fine to add a spring dependency there

@Artur- is there even anything Spring specific in launching the browser ? Looking at the imports, I don't think so.
```java
import org.springframework.context.ConfigurableApplicationContext;
import org.springframework.web.context.support.GenericWebApplicationContext;
```

So why not just move everything related to launching the browser with certain address to the vaadin-dev-server and keep here the logic that digs out what address to open ?

Artur-:
It can for sure be split into two parts: opening a given URL and triggering opening when starting a Spring Boot app. I'll have a look

Artur-:
@mstahv raised another question whether the Spring Boot part should wrap the `run` call like it does right now or if we should move it to a listener so it doesn't show up in the users Application code. What do you think?

pleku:
> @mstahv raised another question whether the Spring Boot part should wrap the `run` call like it does right now or if we should move it to a listener so it doesn't show up in the users Application code. What do you think?

If it doesn't show up in Application code where the call can easily just be removed when someone doesn't like it, then I would think that it should be possible to toggle it off with some configuration. Surplus would be to have it as a visible setting in the "devmodegizmo" but that is not important.

mstahv:
I think also having extra line for the tool after the typical `SpringApplication.run(DemoApplication.class, args);`  call to open the browser would be better than the current solution that makes Vaadin application look like a non-standard Spring Boot application.

Removal of the feature must be easy, I know that the feature is controversial, not everybody likes it.

knoobie:
Personally view from outside: I always remove it, because it is in "my" application code and I have to trust it, that it works as expected and doesn't interfere with my production environment. That's why I exclude fusion and vaadin-dev-server in production builds as well. Adding it inside gizmo / vaadin-dev-server sounds like the better option in the long run.

vaadin-spring/src/main/java/com/vaadin/flow/spring/DevModeBrowserLauncher.java
```diff
@@ -0,0 +1,130 @@
...
+        if (!DevModeBrowserLauncher.isProductionMode(webAppContext)) {
+            String location = getUrl(webAppContext);
+            String outputOnFailure = "Application started at " + location;
+            try {
+                BrowserLauncher.launch(location, outputOnFailure);
+                setLaunched();
+            } catch (Exception | NoClassDefFoundError e) { // NOSONAR
+                // NoClassDefFoundError happens if vaadin-dev-server is not on
+                // the classpath
+                getLogger().info(outputOnFailure);
```
mshabarov, line 77: 1. If `vaadin-dev-server` is not on the classpath, then at least `NoClassDefFoundError` should be logged, or even better be thrown telling to the developer that it's a mistake and this dependency should be added.
2. IMO `Exception` doesn't need to be caught here, otherwise it wouldn't be clear what happened with the application, but instead only  `Application started at ...` would be shown, which gives no idea what was the original error and how to fix it.

vaadin-spring/src/main/resources/META-INF/spring.factories
```diff
@@ -1 +1,2 @@
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=com.vaadin.flow.spring.SpringBootAutoConfiguration,com.vaadin.flow.spring.SpringSecurityAutoConfiguration,com.vaadin.flow.spring.VaadinScopesConfig
+org.springframework.boot.SpringApplicationRunListener=com.vaadin.flow.spring.DevModeBrowserLauncher
```
mshabarov, line 2:
```suggestion
org.springframework.boot.SpringApplicationRunListener=com.vaadin.flow.spring.DevModeBrowserLauncher

```

vaadin-spring/src/test/java/com/vaadin/flow/spring/DevModeBrowserLauncherNoPropertiesTest.java
```diff
@@ -0,0 +1,30 @@
...
+        SpringBootAutoConfiguration.class,
+        SpringSecurityAutoConfiguration.class })
+@TestPropertySource(properties = { "server.port = 1244" })
+public class DevModeBrowserLauncherNoPropertiesTest {
+
+    @Autowired
+    protected GenericWebApplicationContext app;
+
+    @Test
+    public void foo() {
```
mshabarov, line 25:
```suggestion
    public void getUrl_noProperties_givesUrlWithNoContextPathAndUrlMapping() {
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/DevModeBrowserLauncherServletMappingTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockServletContext;
+
+public class DevModeBrowserLauncherServletMappingTest
+        extends DevModeBrowserLauncherNoPropertiesTest {
+
+    @Test
+    public void foo() {
```
mshabarov, line 11:
```suggestion
    public void getUrl_withContextPath_givesUrlWithContextPathAndNoUrlMapping() {
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/DevModeBrowserLauncherVaadinAndServletMappingTest.java
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockServletContext;
+import org.springframework.test.context.TestPropertySource;
+
+@TestPropertySource(properties = { "vaadin.urlMapping=/ui/*" })
+public class DevModeBrowserLauncherVaadinAndServletMappingTest
+        extends DevModeBrowserLauncherNoPropertiesTest {
+
+    @Test
+    public void foo() {
```
mshabarov, line 13:
```suggestion
    public void getUrl_withContextPathAndUrlMapping_givesUrlWithContextPathAndUrlMapping() {
```

vaadin-spring/src/test/java/com/vaadin/flow/spring/DevModeBrowserLauncherVaadinMappingTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.TestPropertySource;
+
+@TestPropertySource(properties = { "vaadin.urlMapping=/ui/*" })
+public class DevModeBrowserLauncherVaadinMappingTest
+        extends DevModeBrowserLauncherNoPropertiesTest {
+
+    @Test
+    public void foo() {
```
mshabarov, line 12:
```suggestion
    public void getUrl_withUrlMapping_givesUrlWithUrlMappingAndNoContextPath() {
```

mshabarov (changes requested):
Nice graceful solution with `SpringApplicationRunListener`!
One comment about exception handling, but otherwise looks good.

What do you think, is it fine to include this feature in v22 (spring add-on is already in beta, sorry for such a long review)?

vaadin-spring/src/main/java/com/vaadin/flow/spring/DevModeBrowserLauncher.java
Artur-: The idea with catching all exceptions is not to bother the user if a browser window couldn't be opened automatically. It is not really a problem with the project and if it doesn't work, you do not want a long exception logged every time you start the server.

Artur-: I also have a feeling that somebody has decided that you should be able to run in development mode without vaadin-dev-server but I do not know why. Am I just wrong @pleku and we should log `NoClassDefFoundError`? (I hope I am)

Artur-:
>What do you think, is it fine to include this feature in v22 (spring add-on is already in beta, sorry for such a long review)?

I think it is quite low risk with the catch all exception handling. We need to update Start, migration instructions and maybe some other places though to avoid double window opening

vaadin-spring/src/main/java/com/vaadin/flow/spring/DevModeBrowserLauncher.java
mshabarov: > The idea with catching all exceptions is not to bother the user if a browser window couldn't be opened automatically. It is not really a problem with the project and if it doesn't work, you do not want a long exception logged every time you start the server.

Okay, I see. So if some exception is thrown and launching the browser fails, then this message is logged giving the URL to developer to click on to open the browser manually. Sounds reasonable.

pleku: It has been possible since v10 to run in not-production-mode but use the production built bundle. It was even requested at some point. Probably not used 99 % of the time but I don't know why we would prevent it now if it is not a significant effort to keep it possible.

mshabarov (approved):

Issue #891, opened 2021-09-01
Spring build fails on fusion security


caalador:
In the afternoon it seems that the mobx has gotten a new version and it is added as `@NpmPackage(value = "mobx", version = "^6.1.5")`
This leads to spring 19, 18, 17 failing validation with
```
[08:08:22]	[Step 2/5] ERROR in frontend/views/view.ts:15:12
[08:08:22]	[Step 2/5] TS2314: Generic type 'IReactionOptions' requires 1 type argument(s).
[08:08:22]	[Step 2/5]     13 |     expression: (r: IReactionPublic) => T,
[08:08:22]	[Step 2/5]     14 |     effect: (arg: T, prev: T, r: IReactionPublic) => void,
[08:08:22]	[Step 2/5]   > 15 |     opts?: IReactionOptions
[08:08:22]	[Step 2/5]        |            ^^^^^^^^^^^^^^^^
[08:08:22]	[Step 2/5]     16 |   ): void {
[08:08:22]	[Step 2/5]     17 |     this.disposers.push(reaction(expression, effect, opts));
[08:08:22]	[Step 2/5]     18 |   }
```

PR #893, opened 2021-09-01
test: fusion test dep with @Npmpackage instead of package.json

haijian-vaadin:
Fixes https://github.com/vaadin/spring/issues/891 (for 17.0)


vaadin-spring-tests/test-spring-security-fusion/src/main/java/com/vaadin/flow/spring/fusionsecurity/Application.java
```diff
@@ -2,13 +2,19 @@
...
 import com.vaadin.flow.theme.Theme;
 
+
 @PWA(name = "Spring Security Helper Test Project", shortName = "SSH Test")
 @SpringBootApplication
 @Theme("spring-security-test-app")
+@NpmPackage(value = "@adobe/lit-mobx",
+            version = "2.0.0-rc.4")
+@NpmPackage(value = "mobx",
+            version = "6.1.5")
```
caalador, line 17: For some reason 17 was using `"mobx": "^6.2.0"` if that has anything on the source compatibility.

caalador (approved):
This brings up the question that why is 18, 19 running an older version?

PR #895, opened 2021-09-01
test: lock mobx to 6.3.3

Artur-:
Reverts vaadin/spring#892

Why would we want to stay on an old version?


haijian-vaadin:
> Reverts #892
> 
> Why would we want to stay on an old version?

Well, it could happen that someday a newer version will break the test in some other way, and CI is suddenly red without code changes. That's why Flow is using locked versions.

The tests here are intended for testing security feature, not testing integration with latest mobx?

Artur-:
Sure but it would make sense to keep using one mobx version across everything. Start projects should be updated to 6.3.3 anyway. So maybe we lock 6.3.3 for now instead?

haijian-vaadin (approved):

PR #896, opened 2021-09-01
test: lock fusion test deps without package.json

haijian-vaadin:
- Add component dependencies for fusion tests via `pom.xml` instead `package.json`
- Remove the `package.json` and `pnpm-lock.yam` files from fusion test modules.
- Lock `mobx` to a newer version: 6.3.3


vaadin-spring-tests/pom.xml
```diff
@@ -137,7 +137,51 @@
                 <artifactId>vaadin-text-field-testbench</artifactId>
                 <version>${vaadin.version}</version>
             </dependency>
-
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-radio-button-flow</artifactId>
+                <version>${vaadin.version}</version>
```
caalador, line 143: The version should not be added as they should come from the managedDependencies in the parent test module.
Anything missing should be added there as managed.

caalador (changes requested):

vaadin-spring-tests/pom.xml
haijian-vaadin: hmm, isn't it what I am doing in this PR? I mean this is the parent test module, I'm just adding the components as managed dependencies here. 

The dependencies are added in the fusion module in the `pom.xml` file in the `test-spring-security-fusion` module

caalador: True. The test module naming confuses. sorry.

caalador (approved):

PR #900, opened 2021-09-14
feat(fusion): add stateless option to VaadinWebSecurityConfigurerAdapter

### LINKED ISSUE #807: Add stateless authentication option to VaadinWebSecurityConfigurerAdapter
We have a `VaadinWebSecurityConfigurerAdapter` which helps with the session based authentications. Would be good to have an option there to help with JWT based stateless authentication, so that when a user wants to use stateless authentication, they could have their own SecurityConfig class by just extending from the `VaadinWebSecurityConfigurerAdapter`.

```
public class SecurityConfig extends VaadinWebSecurityConfigurerAdapter {
    @Override
    protected void configure(HttpSecurity http) throws Exception {
        // @formatter:off
        super.configure(http);

        http
                .csrf().disable()
                .sessionManagement()
                    .sessionCreationPolicy(SessionCreationPolicy.STATELESS);

        setJwtSplitCookieAuthentication(http, "myapp", 3600,
                JWSAlgorithm.HS256);
        setLoginView(http, "/login", "/logout");
        // @formatter:on
    }
}
```

platosha:
Fixes #807


vaadin-spring/pom.xml
```diff
@@ -155,6 +155,14 @@
             <version>${vaadin.flow.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-jose</artifactId>
```
knoobie, line 160: Shouldn't both dependency be optional? Otherwise they gonna leak to all applications

haijian-vaadin: Hi @knoobie, thanks for the comment. Do you see any other problems other than having 2 unnecessary dependencies?
The situation is that we want to enable stateless authentication for Fusion by default. The Spring add-on is shared by both Fusion and Flow, so that if we include these as optional dependencies, then the Fusion app would need to add those dependencies manually, which might be more inconvenient than having 2 unused dependencies for Flow apps. What do you think?

knoobie: Unnecessary and probably conflicting dependencies if flow apps already have some other oauth2-dependencies. Normally flow apps have e.g. `oauth2-client` as dependency but not `oauth2-resource-server`. 

I would suggest to create a separate modules; `spring-flow` and `spring-fusion`. Otherwise changes needed for one framework, leak to the other framework and cause problems for complex customer projects. I'm already exluding `fusion-endpoint` from my flow only apps to reduce transitive dependencies.

knoobie: Example how this could be done with minimal user interruption: 

Split vaadin-spring into two modules; one for each framework. Creating a vaadin-*-spring-boot-starter for each of them and a update the vaadin-spring-boot-starter to have both dependencies. Once this is released; the vaadin-spring-boot-starter could be deprecated and both framework starters could be added independently or combined and be promoted both on e.g. start.spring.io (Vaadin Flow and Vaadin Fusion) (Ping @mstahv because he has done the communication in the past with spring if I remember correctly)

mstahv: @knoobie has a point here. I'd though just make a separate starter for Fusion and keep the old one for Flow (and clean it if there are already obsolete deps for plain Flow users). There isn't that many Fusion apps yet, so it is easy to communicate to them that they need to change the starter dependency. Those using both can add both deps.

vaadin-spring-tests/test-spring-security-fusion-jwt/src/main/java/com/vaadin/flow/spring/fusionsecurityjwt/config/SecurityConfiguration.java
```diff
@@ -47,21 +44,15 @@ protected void configure(HttpSecurity http) throws Exception {
             .sessionManagement()
                     .sessionCreationPolicy(SessionCreationPolicy.STATELESS);
 
-        setJwtSplitCookieAuthentication(http, "statelessapp", 3600,
-                JWSAlgorithm.HS256);
+        setJwtSplitCookieAuthentication(http,
```
haijian-vaadin, line 47: Would it be good to have another overloaded method that takes a string as a secret? so that I can call
```
setJwtSplitCookieAuthentication(http,
                "I72kIcB1UrUQVHVUAzgweE+BLc0bF8mLv9SmrgKsQAk=", "statelessapp");
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/JwtSplitCookieManagementFilter.java
```diff
@@ -1,4 +1,4 @@
-package com.vaadin.flow.spring.fusionsecurityjwt.auth;
+package com.vaadin.flow.spring.security;
```
haijian-vaadin, line 1: License header and Java docs are missing

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/JwtSplitCookieService.java
```diff
@@ -0,0 +1,125 @@
+package com.vaadin.flow.spring.security;
```
haijian-vaadin, line 1: The license header and JavaDoc are missing. 
Also, could we have some unit tests for this class? Looks this is the core of the jwt-based authentication.

haijian-vaadin (changes requested):

haijian-vaadin (changes requested):
The [JwtSecurityUtils](https://github.com/vaadin/spring/blob/ap/feat/stateless-auth/vaadin-spring-tests/test-spring-security-fusion-jwt/src/main/java/com/vaadin/flow/spring/fusionsecurityjwt/JwtSecurityUtils.java) could also be removed?

vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionsecurity/SecurityIT.java
```diff
@@ -36,12 +37,18 @@ public void tearDown() {
...
         });
     }
 
-    private void logout() {
-        if (!$(ButtonElement.class).attribute("id", "logout").exists()) {
+    protected void logout() {
+        ElementQuery<TestBenchElement> mainViewQuery = $("*").attribute("id",
+                "main-view");
+        if (!mainViewQuery.exists() ||
+                mainViewQuery.get(0).$(ButtonElement.class)
```
haijian-vaadin, line 50: why is this condition? I read it as if main doesn't exist or if the logout button exists, then open the mainview?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/SerializedJwtSplitCookieRepository.java
```diff
@@ -0,0 +1,128 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import org.springframework.web.util.WebUtils;
+
+/**
+ * Persists the signed and serialized JWT using a pair of cookies:
+ * "jwt.headerAndPayload" (JS-readable), and "jwt.signature" (HTTP-only).
+ */
+class SerializedJwtSplitCookieRepository {
+    public static final String JWT_HEADER_AND_PAYLOAD_COOKIE_NAME = "jwt.headerAndPayload";
```
haijian-vaadin, line 14: Looks like these 2 constants are used only inside this class, could be private?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinDefaultRequestCache.java
```diff
@@ -52,4 +74,7 @@ private boolean isServiceWorkerInitiated(HttpServletRequest request) {
         return referer != null && referer.endsWith("sw.js");
     }
 
+    public void setDelegateRequestCache(RequestCache delegateRequestCache) {
```
haijian-vaadin, line 77: Do we really need this set method?

haijian-vaadin (changes requested):

vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionsecurity/SecurityIT.java
```diff
@@ -108,13 +115,6 @@ public void navigate_to_private_view_prevented() {
         assertLoginViewShown();
     }
 
-    @Test
-    public void navigate_to_admin_view_prevented() {
-        open("");
-        navigateTo("admin", false);
```
haijian-vaadin, line 114: why this test was removed?

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/VaadinWebSecurityConfigurerAdapter.java
```diff
@@ -245,10 +247,63 @@ protected void setLoginView(HttpSecurity http,
...
+     *            the http security from {@link #configure(HttpSecurity)}
+     * @param secretKey
+     *            the secret key for encoding and decoding JWTs, must use a
+     *            {@link MacAlgorithm} algorithm name
+     * @param issuer
+     *            the issuer JWT claim
+     * @throws Exception
+     *             if something goes wrong
+     */
+    protected void setJwtSplitCookieAuthentication(HttpSecurity http,
```
haijian-vaadin, line 269: Is the API name a bit too technical? would it be better with sth like `setStatelessAuthentication()`? so that we also don't bind the API to the implementation details.

haijian-vaadin (changes requested):

vaadin-spring/pom.xml
haijian-vaadin: Thanks for the suggestion, marked those as optional dependencies.

haijian-vaadin (approved):

PR #903, opened 2021-09-17
test: use @vaadin/fusion-frontend for fusion imports

haijian-vaadin:
Use `@vaadin/fusion-frontend` for fusion imports.

### DIFF
--- a/vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionform/FusionFormIT.java
+++ b/vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionform/FusionFormIT.java
@@ -7,6 +7,7 @@ import com.vaadin.flow.testutil.ChromeBrowserTest;
 
 import org.junit.After;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class FusionFormIT extends ChromeBrowserTest {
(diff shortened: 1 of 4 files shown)


vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionform/FusionFormIT.java
```diff
@@ -66,6 +67,7 @@ public void save_backend_loaded_empty_values_for_required_fields_no_runtime_erro
     }
 
     @Test
+    @Ignore
     // https://github.com/vaadin/fusion/issues/13
     public void no_validation_error_when_clearing_number_field() {
```
caalador, line 72: Why would this fail if only the package has been renamed?

caalador (changes requested):

vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionform/FusionFormIT.java
haijian-vaadin: It's due to a missing mix in the new fusion package. I will port the fix, make a release of fusion-frontend, update flow to use the new version, and make another or in spring to reenable it. It will take a while, so thought let's make the Spring green first.

caalador: I'm ok with that if there is a ticket about enabling the test so it's not forgotten.

haijian-vaadin: fair enough, made a ticket https://github.com/vaadin/spring/issues/904.

caalador (approved):

PR #935, opened 2021-11-17
Make vaadin.blacklisted-packages specifiable with packages

tarekoraby:
Unlike `vaadin.whitelisted-packages` which can be specified with both '.' and '/', `vaadin.blacklisted-packages` only works with '/'. 
This is suboptimal, as rightly noted in this [issue](https://github.com/vaadin/starters/issues/34), because packages use . not /.

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -558,7 +558,7 @@ public class VaadinServletContextInitializer
             neverScan = Collections.emptyList();
         } else {
             neverScan = Arrays.stream(neverScanProperty.split(","))
-                    .map(onlyFolder -> onlyFolder.replace('.', '/').trim())
+                    .map(pkg -> pkg.replace('.', '/').trim())
                     .collect(Collectors.toList());
         }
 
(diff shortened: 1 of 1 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -558,7 +558,8 @@ public VaadinServletContextInitializer(ApplicationContext context) {
             neverScan = Collections.emptyList();
         } else {
             neverScan = Arrays.stream(neverScanProperty.split(","))
-                    .map(String::trim).collect(Collectors.toList());
+                    .map(onlyFolder -> onlyFolder.replace('.', '/').trim())
```
Artur-, line 561: `onlyFolder` - what does that refer to?

tarekoraby: To use the folder instead of package (to be symmetrical to how `onlyPackages` is used with `vaadin.whitelisted-packages`)

tarekoraby: Though I know realize that "only" refers to the inclusion criterion of whitelisted packages.

tarekoraby: Changed to `pkg `

Artur- (approved):

PR #939, opened 2021-11-19
feat: introduce JwtClaimsSource

platosha:
Fixes vaadin/fusion#46

### DIFF
--- a/vaadin-spring-tests/test-spring-security-fusion/src/main/java/com/vaadin/flow/spring/fusionsecurity/SecurityConfig.java
+++ b/vaadin-spring-tests/test-spring-security-fusion/src/main/java/com/vaadin/flow/spring/fusionsecurity/SecurityConfig.java
@@ -2,6 +2,8 @@ package com.vaadin.flow.spring.fusionsecurity;
 
 import javax.crypto.spec.SecretKeySpec;
 import java.util.Base64;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -22,6 +24,8 @@ import org.springframework.security.oauth2.jose.jws.JwsAlgorithms;
 import com.vaadin.flow.spring.fusionsecurity.data.UserInfo;
 import com.vaadin.flow.spring.fusionsecurity.data.UserInfoRepository;
 import com.vaadin.flow.spring.security.VaadinWebSecurityConfigurerAdapter;
+import com.vaadin.flow.spring.security.stateless.DefaultJwtClaimsSource;
+import com.vaadin.flow.spring.security.stateless.VaadinStatelessSecurityConfigurer;
 
 @EnableWebSecurity
 @Configuration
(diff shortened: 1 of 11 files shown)


vaadin-spring-tests/test-spring-security-fusion/src/main/java/com/vaadin/flow/spring/fusionsecurity/SecurityConfig.java
```diff
@@ -55,11 +58,23 @@ protected void configure(HttpSecurity http) throws Exception {
         setLoginView(http, "/login");
 
         if (stateless) {
-            setStatelessAuthentication(http,
-                    new SecretKeySpec(Base64.getUrlDecoder().decode(
+            // setStatelessAuthentication(http,
```
haijian-vaadin, line 61: could remove this commented code?

haijian-vaadin (commented):
I got a feeling we could remove the extra layer of JwtClaminsSource? instead could use a `JwtClaimAccessor`?

vaadin-spring-tests/test-spring-security-fusion/src/main/java/com/vaadin/flow/spring/fusionsecurity/SecurityConfig.java
platosha: Temporarily commented for testing. Maybe I should rather put it back, we probably want to keep `setStatelessAuthentication` method tested.

platosha:
> I got a feeling we could remove the extra layer of `JwtClaimsSource`? instead could use a `JwtClaimAccessor`?

Not sure if it’s nice in practice. We need a converter from the authentication data to the JWT claims set.  `JwtClaimAccessor` basically is meant to represent the JWT claims set itself, so it’s not usable as a converter.

One way about it is to support a generic `Function<Authentication, JwtClaimAccessor>` type of converter. I have an impression that use case specific types are a better fit though, especially for dependency injection that is type-dependant.

platosha:
My plan here is to replace `JwtClaimAccessor` with even more specific [`JwtClaimsSet`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimsSet.java) now that it is released.

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/JwtSecurityContextRepository.java
```diff
@@ -133,13 +139,17 @@ private String encodeJwt(Authentication authentication)
             return null;
         }
 
-        final Date now = new Date();
+        JwtClaimAccessor claims = jwtClaimsSource.get(authentication);
```
haijian-vaadin, line 142:
```suggestion
        JwtClaimAccessor claimAccessor = jwtClaimsSource.get(authentication);
```

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/DefaultJwtClaimsSource.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.security.oauth2.jwt.JwtClaimAccessor;
+
+/**
+ * A {@link JwtClaimsSource} implementation that creates a set with the {@code
+ * "sub"} and {@code "roles"} JWT claims using the authentication principal name
+ * and roles respectively.
+ */
+public final class DefaultJwtClaimsSource implements JwtClaimsSource {
```
haijian-vaadin, line 31: I think it's better to make it not final, so that as a developer, I can create my own JwtClaimSource by extending the default one

haijian-vaadin (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/security/stateless/DefaultJwtClaimsSource.java
platosha:
```suggestion
public class DefaultJwtClaimsSource implements JwtClaimsSource {
```

haijian-vaadin (approved):

haijian-vaadin (approved):

haijian-vaadin (changes requested):
Per discussion, let's wait until #940 and #944 are merged, then we use `JwtClaimSet` instead `JwtClaimAccessor`

Artur-:
This should go into https://github.com/vaadin/flow

PR #942, opened 2021-11-24
Revert "Fix tests"

Artur-:
Reverts vaadin/spring#937

It is not reasonable to think that everybody should do this in their tests


Artur-:
Should pass once https://github.com/vaadin/flow/pull/12437 is merged

platosha (approved):

PR #940, opened 2021-11-22
chore: update spring-boot to 2.6.2

Artur-:

PR #944, opened 2021-11-24
fix: Spring Boot 2.6 compatibility

### LINKED ISSUE #941: Does not work with Spring Boot 2.6
Seems like Spring Boot 2.6 removed `ResourceProperties` which is used in some hacky code in https://github.com/vaadin/spring/blob/dee6efa77a5cb4455ebcc244b4a25778b2eeefa4/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java#L136

Artur-:
Fixes #941

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
@@ -47,6 +47,8 @@ public class SpringVaadinServletService extends VaadinServletService {
 
     private final Registration serviceDestroyRegistration;
 
+    private static final String SPRING_BOOT_RESOURCES_CLASS = "org.springframework.boot.autoconfigure.web.WebProperties.Resources";
+
     /**
      * Creates an instance connected to the given servlet and using the given
      * configuration with provided application {@code context}.
@@ -133,7 +135,7 @@ public class SpringVaadinServletService extends VaadinServletService {
             return null;
         }
         for (String prefix : context.getBean(
-                org.springframework.boot.autoconfigure.web.ResourceProperties.class)
+                org.springframework.boot.autoconfigure.web.WebProperties.Resources.class)
                 .getStaticLocations()) {
             Resource resource = context.getResource(getFullPath(path, prefix));
             if (resource != null) {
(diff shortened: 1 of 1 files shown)


ZheSun88:
do we need to update https://github.com/vaadin/spring/blob/master/pom.xml#L47 to spring-boot 2.6.0?

Artur-:
This PR should be compatible with Spring Boot 2.4+. Maybe we should still update the version at the same time

Artur-:
This fix could go into V22, we probably shouldn't go to Spring Boot 2.6 for that at this point

Artur-:
Let's do the version update in #940

knoobie:
The FQN defined in isSpringBootConfigured is probably faulty as well

lmorocz:
And there is an extra `org.springframework.boot.autoconfigure.web.ResourceProperties.class` in the `isSpringBootConfigured` method as well.

Artur-:
Simplified

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
```diff
@@ -22,6 +22,8 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.boot.autoconfigure.web.WebProperties.Resources;
```
knoobie, line 26: Is this class to be expected to work with plain spring or just spring boot?

platosha (dismissed):

haijian-vaadin:
could you @Artur- help to fix the sonar issue, let's get this merged soon

vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringVaadinServletService.java
Artur-: It is supposed to work with Spring. Seems like there are no proper tests for plain Spring projects in this repository

Artur-: Let's keep the untested code then and hope that it might work with plain Spring..

knoobie: I would like to have a more spring boot oriented implementation as well.. but I've seen multiple PRs in this repo in the past where explicitly a more complex route was taken to reduce the coupling with spring boot.. that's why I wanted to mention it

mstahv:
I wouldn't care about Sonar issues at all, but Spring Boot 2.6.0 support for 22 would be cool ;-)

knoobie:
^ for 14.6 as well 😀

mstahv:
For 14 series as well yeah 👍

Artur-:
Seems to work now, just waiting for a review

caalador (approved):

joheriks:
> Seems like this is already in the 19.0 branch
There was no `19.0` branch at that point, created it after this commit.

PR #951, opened 2021-11-30
Add whitelisted-packages' docs URL to logs

tarekoraby:
When scanning is slow, specify the URL in the docs in which whitelisted-packages feature is discussed. 

This should be (1) easier for the developer than having to navigate to the relevant section themselves, and (2) more future-proof in case the structure of the docs changes (because usually there are redirects in place when the docs' structure is altered).

### DIFF
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -457,7 +457,7 @@ public class VaadinServletContextInitializer
                     .getProperty("vaadin.whitelisted-packages") == null) {
                 getLogger().info(
                         "Due to slow search it is recommended to use the whitelisted-packages feature to make scanning faster.\n\n"
-                                + "See 'Flow / Integrations / Using Vaadin with Spring / Vaadin Spring Configuration' topic 'Special configuration parameters' at https://vaadin.com/docs for more detailed information.");
+                                + "See the whitelisted-packages section in the docs at https://vaadin.com/docs/latest/flow/integrations/spring/configuration#special-configuration-parameters.");
             }
 
             try {
--- a/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
+++ b/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
@@ -457,7 +457,7 @@ public class VaadinServletContextInitializer
                     .getProperty("vaadin.whitelisted-packages") == null) {
                 getLogger().info(
                         "Due to slow search it is recommended to use the whitelisted-packages feature to make scanning faster.\n\n"
-                                + "See the whitelisted-packages section in the docs at https://vaadin.com/docs/latest/flow/integrations/spring/configuration#special-configuration-parameters.");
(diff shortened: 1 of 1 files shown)


vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
```diff
@@ -457,7 +457,7 @@ public void failFastContextInitialized(ServletContextEvent event)
                     .getProperty("vaadin.whitelisted-packages") == null) {
                 getLogger().info(
                         "Due to slow search it is recommended to use the whitelisted-packages feature to make scanning faster.\n\n"
-                                + "See 'Flow / Integrations / Using Vaadin with Spring / Vaadin Spring Configuration' topic 'Special configuration parameters' at https://vaadin.com/docs for more detailed information.");
+                                + "See the whitelisted-packages section in the docs at https://vaadin.com/docs/latest/flow/integrations/spring/configuration#special-configuration-parameters.");
```
caalador, line 460:
```suggestion
                                + "See the 'Special configuration parameters section' in the docs at https://vaadin.com/docs/latest/flow/integrations/spring/configuration#special-configuration-parameters.");
```
At the moment it is the `special configuration parameters` section in the document. I was not setting a link as the version is in many cases wrong for the link even if it should perhaps be the same for v14 and v20+
@joheriks what is your take on using the link vs giving where the documentation exists at the moment.

caalador (changes requested):

vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinServletContextInitializer.java
tarekoraby: Yes, the feature is discussed under the `Special configuration parameters` section in the docs. But the name of the section is not that important IMO. What's more important here is to make it clear to the developer that if they want to configure `whitelisted-packages`, then they should follow the provided link.

> I was not setting a link as the version is in many cases wrong for the link even if it should perhaps be the same for v14 and v20+

The link I added should be always valid for `vLatest` (and the content, as you rightly mentioned, should be identical across platform versions).

joheriks: Links to docs are useful but very specific ones, such as to internal anchors, are prone to breakage. So here I think we would be better off linking to https://vaadin.com/docs/latest/flow/integrations/spring/configuration, which should remain fairly stable and be platform version independent.

tarekoraby: Hash fragments are not sent to the server, they are handled by the client. This means that if the fragment (i.e. the part after the `#`) gets changed in the docs for some reason, the client will still be routed to the `/configuration` page (so there is no downside to adding the fragment). 

You can test by trying [/configuration#noexistent-fragment](https://vaadin.com/docs/latest/flow/integrations/spring/configuration#noexistent-fragment).

caalador:
```suggestion
                                + "See the whitelisted-packages section in the docs at https://vaadin.com/docs/latest/flow/integrations/spring/configuration#special-configuration-parameters");
```
No dot at the end at least as it is not clear that it shouldn't be copied.

caalador (approved):

PR #975, opened 2021-12-10
Test Spring 2.6.2-SNAPSHOT

Artur-:


vaadin-spring-tests/test-spring-security-fusion/src/test/java/com/vaadin/flow/spring/fusionsecurity/SecurityIT.java
```diff
@@ -329,8 +329,7 @@ protected void refresh() {
     }
 
     private void assertForbiddenPage() {
-        assertPageContains(
-                "There was an unexpected error (type=Forbidden, status=403).");
+        assertPageContains("HTTP Status 403");
```
tarekoraby, line 332: I guess that's why the automated build of `2.6.2-SNAPSHOT` on bender is failing. Should we commit a change to this test such that it passes if either of the two strings is true (thereby it would pass for both `2.6.2-SNAPSHOT` and `2.5.7`?

PR #987, opened 2021-12-21
chore: Upgrade to Spring Boot 2.5.8

Artur-:


Artur-:
Let's do #940 instead as it now seems to work

PR #1001, opened 2022-01-19
fix: Remove dependency to outdated oauth2 library

Artur-:


Artur-:
This should go into https://github.com/vaadin/flow

PR #1043, opened 2022-10-20
chore: Add license checker 1.10.1

mshabarov:


caalador (commented):
Should probably also update the flow version in pom as now validation has run against `2.7-SNAPSHOT` (from the test run `com.vaadin : flow-server : jar : 2.7-20221020.070407-672`)

Also wondering why doesn't the trasitive dependency from flow work here if it works with master?

mshabarov:
> Should probably also update the flow version in pom as now validation has run against `2.7-SNAPSHOT` (from the test run `com.vaadin : flow-server : jar : 2.7-20221020.070407-672`)
> 

Sure. I've tested locally with both flow versions and left 2.7 unchanged.

> Also wondering why doesn't the trasitive dependency from flow work here if it works with master?

Can't say. This is not the first time I observe this maven behaviour.

caalador (approved):

PR #1046, opened 2022-11-25
chore: Upgrade to Spring Boot 2.7.7

Artur-:
Spring Boot 2.5 and 2.6 are end of life

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
 
         <!-- These are typically overridden with BOMs -->
         <vaadin.flow.version>9.0-SNAPSHOT</vaadin.flow.version>
-        <spring-boot.version>2.5.12</spring-boot.version>
+        <spring-boot.version>2.7.6</spring-boot.version>
 
         <!-- Additional manifest fields -->
         <Vaadin-License-Title>Apache License 2.0</Vaadin-License-Title>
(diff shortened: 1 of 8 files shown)


mcollovati:
okhttp client version managed by spring conflicts with the one supported by the selenium version used in testbench 7.0.3
Upgrading testbench is not an option because base tests classes in Flow 9.0 uses 7.0.3 and there will be compilation errors.
Solution may be adding an entry for `com.squareup.okhttp3:okhttp:3.11.0`  in dependency management section oif `vaadin-spring-tests` module.

mcollovati:
@Artur- is it ok to pin `okhttp3` in `vaadin-spring-tests module`?

mcollovati:
@Artur- do you mind taking a look at the additional changes needed to upgrade to Spring Boot 2.7.7?

Artur-:
Is the `spring.factories´ change the only one that explicitly makes it incompatible with earlier versions? If so, does it make sense to do it or just leave it use the deprecated variant "forever"?

mcollovati:
> Is the `spring.factories´ change the only one that explicitly makes it incompatible with earlier versions? If so, does it make sense to do it or just leave it use the deprecated variant "forever"?

It seems so, other changes are all related to our tests.
I'll revert the `spring.factories` change

Artur-:
👍  but I cannot approve my own PR

mcollovati:
Thanks. I can approve it

mcollovati (approved):

PR #1087, opened 2024-08-26
feat: Finish update to new Spring versions

thevaadinman:
* Updates Vaadin to 7.7.44
* Updates Spring to 5.3.39
* Updates Spring Boot to 2.7.18
* Introduces Spring Security 5.7.11
* Updates javax.api to 4.1.0
* Fixes build scripts

TODO: Introduce static file path logic to Vaadin 7's VaadinServlet for Vaadin 7.7.45

### DIFF
--- a/vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/VaadinAutoConfigurationWithCustomNavigatorTest.java
+++ b/vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/VaadinAutoConfigurationWithCustomNavigatorTest.java
@@ -37,8 +37,7 @@ import com.vaadin.spring.server.AbstractSpringUIProviderTest;
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration
 @WebAppConfiguration
-// make sure the context is cleaned
-@DirtiesContext
+@DirtiesContext     // make sure the context is cleaned
 public class VaadinAutoConfigurationWithCustomNavigatorTest
         extends AbstractSpringUIProviderTest {
 
@@ -75,7 +74,7 @@ public class VaadinAutoConfigurationWithCustomNavigatorTest
     public void testNavigatorCustomized() {
         // this sets up the UI scope
         TestUI ui = createUi(TestUI.class);
-        Assert.notNull(ui);
+        Assert.notNull(ui, "UI should not be null");
         Assert.isInstanceOf(MyNavigator.class,
                 applicationContext.getBean(SpringNavigator.class),
                 "Vaadin Navigator is not correctly overridden");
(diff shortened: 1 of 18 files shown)


TatuLund (approved):

vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/VaadinAutoConfigurationWithCustomNavigatorTest.java
```diff
@@ -37,8 +37,7 @@
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration
 @WebAppConfiguration
-// make sure the context is cleaned
-@DirtiesContext
+@DirtiesContext     // make sure the context is cleaned
```
Ansku, line 40: Too much whitespace.

vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/internal/VaadinServletConfigurationUIPathTest.java
```diff
@@ -15,13 +15,12 @@
...
 package com.vaadin.spring.boot.internal;
 
-import static org.junit.Assert.assertTrue;
-
 import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.junit.Assert;
```
Ansku, line 23: Wrong direction, would be better to switch to use the method imports everywhere instead.

vaadin-spring/src/main/java/com/vaadin/spring/internal/UIScopeImpl.java
```diff
@@ -229,6 +249,9 @@ void removeBeanStore(UIID uiid) {
 
         void destroy() {
             LOGGER.trace("Destroying [{}]", this);
+            session.accessSynchronously(() -> {
+                session.setAttribute(UIStore.class, null);
```
Ansku, line 253: Why is this done here instead of directly after that `uiStore.destroy()` call? Is it possible that that new `cleanupSession` method doesn't get called before this one is? If so, is it possible that `uiStore.destroy()` never gets called, and is it a problem if that were to happen?

thevaadinman: That comes from the Vaadin 8 version, https://github.com/vaadin/spring/blob/3.0/vaadin-spring/src/main/java/com/vaadin/spring/internal/UIScopeImpl.java#L255

I don't actually know if that is a problem or not, and at the very least just looking at the code I fail to see the actual issue; it seems to work in the V8 version.

vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/VaadinAutoConfigurationWithCustomNavigatorTest.java
thevaadinman: Cosmetic; can be fixed but is not high priority.

vaadin-spring-boot/src/test/java/com/vaadin/spring/boot/internal/VaadinServletConfigurationUIPathTest.java
thevaadinman: Cosmetic; can be fixed but is not high priority.

PR #1089, opened 2025-02-03
fix: Try to compile with Java 11

thevaadinman:


thevaadinman:
This was a test, and the test failed successfully. :)

